PR #19, opened 2022-06-01
ilog function and test

### LINKED ISSUE #15: Add `ilog()` function because why not

nadinehussein:
closes #15

### DIFF
--- a/R/log.R
+++ b/R/log.R
@@ -16,3 +16,21 @@
 `log<-` <- function(x, value) {
   exp(value)
 }
+
+#' Log Transformation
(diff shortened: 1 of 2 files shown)


R/log.R
```diff
@@ -16,3 +16,21 @@
 `log<-` <- function(x, value) {
   exp(value)
 }
+
+#' Log Transformation
```
joethorley, line 20: This needs remaining the title to Inverse Log Transformation

joethorley (changes requested):

PR #21, opened 2022-06-01
inv_odds functions and tests

### LINKED ISSUE #12: Add `invodds()` and `odds<-` to convert odds to probability

nadinehussein:
Closes #12.

Only caveat is can't convert from `odds(1)` back with `inv_odds` due to `odds(1)` returning Inf.


joethorley:
thanks @nadinehussein - that's great

PR #50, opened 2023-05-25
Add `exp2()` and `exp10()` functions

### LINKED ISSUE #44: Add `exp10(x)` and `exp2(x)` functions
For `10^x` and `2^x`.

Should have the same behaviour as exp(x).

nehill197:
Add transformation functions for 2^x and 10^x

Fixes #44


joethorley (changes requested):
looks good but we should either drop checks or use `chk::chk_numeric(x)`

nehill197:
@joethorley is this better? I need these additional lines to keep the same behaviour as `exp()`:
- NAs pass through
- error if x is NULL or not numeric

aylapear:
Update the docs 
- remove details tag
- update details tag with value 
- add a second example
- remove examples not of the function being written
- remove output of examples
- Full stops for the description sentence/paragraph, ie end sentences with a period
- Title case for Title tag, ie important words are capitalized and no period at end 

Update tests
- split tests into their own test_that block with a descriptive name
- test various types (strings, integers, real)

aylapear (changes requested):
See comment in pull request for requested changes

joethorley (approved):

aylapear (changes requested):
Two minor comments: 
- the return value of the two functions is written slightly differently, it would be nice to have them aligned (A numeric atomic object with the value of 2^x. vs A numeric atomic object.)
- The title of each function should be in title case so "Exponential transformation of base 2" should be "Exponential Transformation of Base 2"

Thanks for splitting up the tests, that looks good!

joethorley (approved):
Excellent!

joethorley (approved):

aylapear (approved):

PR #52, opened 2023-05-26
Add `log2()<-`, `log10()<-`, `ilog2()`, and `ilog10()` functions

nehill197:
To allow RHS and inverse evaluation of `2^x` and `10^x`.
These functions rely on the `exp2()` and `exp10()` functions.


joethorley (approved):

aylapear (changes requested):
- Review the parameter definitions in the params file to see if they are applicable for these functions.
- Split up the tests a bit further so each block is self contained to a single idea. So for example the edge cases block where it tests NULL, NA and Inf I would split those into their own test block as they are testing separate issues.

joethorley (approved):

aylapear (approved):

PR #55, opened 2023-07-27
Skew normal residual functions

nehill197:
Add `dev_skewnorm`, `res_skewnorm`, `log_lik_skewnorm`, and `ran_skewnorm` functions for use with models employing the skew normal distribution.


joethorley (approved):
Looks great - merge away

PR #58, opened 2024-05-15
Add sensitivity functions to produce vector of new pars that vary sd without changing mean

nehill197:
References issue #59


nehill197:
extras is dependent on the `sn` package, which in turn is dependent on the `Matrix` package, which now requires R >= 4.4.0 (though the actions passed on R 4.3.0). I changed the R dependency to 4.3.0, and updated the R-CMD-check actions from the "overkill" set that included the windows on R 4.1 and the older ubuntu versions to the "check standard" version.

nehill197:
The binomial family of distributions don't seem to have a clear solution to adjust the parameters to expand the sd.
I also haven't included a function for the zero-inflated Poisson distribution because it's proving difficult.

nehill197:
This pull request also includes a handful of styler and lintr suggestions to some of the other functions/tests/vignettes.

joethorley (approved):
looks great!

PR #65, opened 2024-07-22
Refactor `log_lik_beta_binom()` and use styler on package

nehill197:
I sped up the `log_lik_beta_binom()` function through some minor refactoring.
This didn't produce significant differences in run time when running posterior predictive checks, but I kept the changes.

I wrote a non-vectorized C++ function to calculate the beta binomial log-likelihood, and used this in the call to `optimize()` within `dev_beta_binom()`.
Microbenchmarking shows that optimizing using the C++ function is much faster than the previous version that optimized using the `extras::log_lik_beta_binom()` function.

image


joethorley (changes requested):
Good work on such a great speed gain. This is something we would work on with Kirill in August when you come and join us. To facilitate the conversation could you modify this PR so that it includes all the formatting updates but not the Rcpp ones and merge it in. 

And then do a separate PR with just the Rcpp related changes so that they are easier to see.

Awesome work!

joethorley (approved):

PR #70, opened 2024-08-19
Skip tests if suggested packages aren't installed

nehill197:
The `aods3` package is no longer available on CRAN and cannot be installed. Only one test relies on it that checks the beta-binomial log-likelihood.

The CRAN team said to ensure tests that use packages in suggests should be used conditionally.
I used the `testthat::skip_if_not_installed()` function to skip the `aods3` test, as well as another test that requires the `MASS` package.


nehill197:
The mac R-CMD-check is failing because aods3 is not available to be installed; the other actions use a cached version.
`aods3` removed from CRAN today! (08-19-24)
Emailed the maintainer to see if they are planning to fix.

nehill197:
Decided to remove dependency on `aods3`; replaced the test that relied on it with a snapshot test.

joethorley (approved):
looks great

PR #71, opened 2024-08-20
Skip skewnorm tests if `sn` package not installed

nehill197:
R-hub actions indicated these tests should also be conditional!


joethorley (approved):
great

Issue #72, opened 2024-08-20
Replace  `requireNamespace()` with `rlang::check_installed()`


joethorley:
For example

```
  if (!requireNamespace("sn", quietly = TRUE)) {
    stop(
      "Package \"sn\" must be installed to use this function.",
      call. = FALSE
    )
  }
```

```
rlang::check_installed("sn")
```

PR #81, opened 2024-09-17
Upkeep package

nehill197:
Perform upkeep on package #77 and replace `requireNamespace()` with `rlang::check_installed()` or `rlang::is_installed()` #72


joethorley (approved):
Looks great

PR #93, opened 2026-01-19
- Add `log_lik_exp()`

### LINKED ISSUE #90: Add log_lik_exp

joethorley:
Fixes #90


R/log-lik.R
nehill197: I think that the mean parameterization makes sense within the context of what we've done with the rest of the package, but I am a little concerned about myself and others forgetting that it's the mean instead of the rate when filling in the sensitivity part of the new expression, and getting wonky sensitivity results.

nehill197 (commented):
Just a little unsure about the parameterization - let's discuss?

R/log-lik.R
joethorley: I've switched it to the rate parameterization based on your and @sebdalgarno's  recommendation

nehill197 (approved):
Great! Thanks Joe

PR #100, opened 2026-04-22
allow one-sided p-values

StefanoMezzini:


StefanoMezzini:
we may want to also add a check that returns a warning if `length(x)` is too small and thus the `+1` to move away from zero has a large effect (e.g., `(0 + 1) / (10 + 1) = 0.91`).

aylapear:
It would be good to get the actions working before making any changes to the package so we can ensure the changes don't break anything.

aylapear:
Tests should be added for the changes.

joethorley (changes requested):
As one_sided can be directional I'm wondering if we shouldn't go with 

sided = "two"

with "left" or "right" as alternative options because ideally arguments should be orthogonal ie independent of other argument values.

joethorley:
Another package called the argument direction  but I prefer sided. It was more relaxed on how the argument can be specified but this is as nightmare for refactoring code in the future.

Screenshot 2026-04-22 at 19 13 12

joethorley:
Great question on the warning - my feeling is that it's better not to issue one because the function will be called for each parameter in a model potentially resulting in 10s or 100s of warnings and because ultimately its up to the user to know that if they have a model with < 100 MCMC samples any estimates or derived quantities will not be reliable.

StefanoMezzini:
RE: testing -- I had written some tests, but they ended up going into a different branch (2sided-pvalues) because I had misnamed the branch initially, and renaming it only changed it for the first commit. I've merged the branch into this one.

StefanoMezzini:
I agree that merging the two arguments is probably a good idea, but I think `sided` implies a logical argument. I think `side` with options `c("left", "right", "both")` would avoid this confusion.

R/pvalue.R
```diff
@@ -15,10 +16,18 @@
...
+#' pvalue(x) # should be 0.05 * 2
+#' pvalue(x, side = "left") # should be 0.95
+#' pvalue(x, side = "right") # should be 0.05
+pvalue <- function(x, threshold = 0, side = "both", na_rm = FALSE) {
   chk_numeric(x)
   chk_number(threshold)
+  chk_character(side)
+  if (! side %in% c("left", "right", "both")) {
+    stop("`side` must be \"left\", \"right\", or \"both\".")
+  }
```
joethorley:
```suggestion
chk_subset(side, c("left", "right", "both"))
```

R/pvalue.R
```diff
@@ -15,10 +16,18 @@
...
-pvalue <- function(x, threshold = 0, na_rm = FALSE) {
+#' x <- rnorm(1e6, qnorm(0.05, lower.tail = TRUE))
+#' hist(x); abline(v = 0, col = "red", lwd = 2, lty = 2)
+#' pvalue(x) # should be 0.05 * 2
+#' pvalue(x, side = "left") # should be 0.95
+#' pvalue(x, side = "right") # should be 0.05
+pvalue <- function(x, threshold = 0, side = "both", na_rm = FALSE) {
   chk_numeric(x)
   chk_number(threshold)
+  chk_character(side)
```
joethorley:
```suggestion
  chk_string(side)
```

R/pvalue.R
```diff
@@ -30,11 +39,24 @@ pvalue <- function(x, threshold = 0, na_rm = FALSE) {
...
+    s1 <- sum(x < threshold)
+    s2 <- sum(x > threshold)
+    s <- min(s1, s2)
+    s <- s * 2 # two sided p-value
+    s <- s + n - s1 - s2 # include threshold values
+    p <- (s + 1) / (n + 1) # avoid pvalues of 0
+    return(p)
+  } else {
+    if (side == "left") {
+      s <- sum(x <= threshold) # include threshold values
```
joethorley:
```suggestion
      s <- sum(x < threshold) # include threshold values
```

I think we want to exclude a match?

R/pvalue.R
```diff
@@ -30,11 +39,24 @@ pvalue <- function(x, threshold = 0, na_rm = FALSE) {
...
+    } else {
+      s <- sum(x >= threshold) # include threshold values
```
joethorley:
```suggestion
      s <- sum(x > threshold) # include threshold values
```

joethorley (changes requested):
Looks good I've got a few edits which I think make sense

R/pvalue.R
StefanoMezzini: ah, I was wondering if something like that existed. I should have checked!

R/pvalue.R
```diff
@@ -15,10 +16,18 @@
...
+#' x <- rnorm(1e6, qnorm(0.05, lower.tail = TRUE))
+#' hist(x); abline(v = 0, col = "red", lwd = 2, lty = 2)
+#' pvalue(x) # should be 0.05 * 2
+#' pvalue(x, side = "left") # should be 0.95
+#' pvalue(x, side = "right") # should be 0.05
+pvalue <- function(x, threshold = 0, side = "both", na_rm = FALSE) {
   chk_numeric(x)
   chk_number(threshold)
+  chk_character(side)
+  if (! side %in% c("left", "right", "both")) {
```
aylapear: I think you can do this check with `chk::chi_subset()`

R/pvalue.R
StefanoMezzini: no, we want to include the match because the p-value is for values "as extreme or more". We also include the threshold values in the two-sided calculation above.

R/pvalue.R
StefanoMezzini: see comment above

R/pvalue.R
```diff
@@ -30,11 +39,24 @@ pvalue <- function(x, threshold = 0, na_rm = FALSE) {
...
+  } else {
+    if (side == "left") {
+      s <- sum(x <= threshold) # include threshold values
+    } else {
+      s <- sum(x >= threshold) # include threshold values
+    }
+    n <- length(x)
+    p <- (s + 1) / (n + 1) # avoid pvalues of 0
+    return(p)
+  }
```
aylapear, line 56: Since you are doing three conditional statements I would think about switching it to two additional `else if` statements instead of an else and then if and else inside it. The `else if` statements are more readable because they show the condition. I don't think you need the return statement at the end of the loop either.

tests/testthat/test-pvalue.R
```diff
@@ -20,4 +20,53 @@ test_that("pvalue", {
...
   expect_equal(pvalue(c(-1, 1, 1, 1, NA)), NA_real_)
   expect_equal(pvalue(c(-1, 1, 1, 1, NA), na_rm = TRUE), 3 / 5)
+
+  expect_equal(round(pvalue(rnorm(1e7, mean = qnorm(0.95))), 2),
+               0.05 * 2)
+  expect_equal(round(pvalue(rnorm(1e7, mean = qnorm(0.05))), 2),
+               0.05 * 2)
+})
+
+test_that("pvalue() returns the correct one-sided p-values", {
```
aylapear, line 30: Since you are using random generating functions is there any kind of seed that should be set in the tests? The `withr` does have a seed setting function. If tests are passing after multiple runs it may not be needed but good to thing if changes in the future could cause issues here to make it more robust.

aylapear (changes requested):

R/pvalue.R
StefanoMezzini: i split them like this because some of the code for the left and right tails repeats, but i can split them since it's harder to read.

I didn't know the tidyverse suggested `return()` only be used for early exits. I'll remove it.

StefanoMezzini: fixed in 90f3da4a5e823d45288d0e085f8c4e304ce97756 and bdca167ed5906b4b343abade4d69aef00ae1e914, respectively

tests/testthat/test-pvalue.R
StefanoMezzini: I could set it up, but the rounding should make it robust enough that an error is extremely unlikely, given the sample size of 1e7 and the rounding to 2 decimals

R/pvalue.R
StefanoMezzini: fixed in commit 752a90e903c64cabcb700343d068141055d1894d

R/pvalue.R
```diff
@@ -15,10 +16,16 @@
...
+#' x <- rnorm(1e6, qnorm(0.05, lower.tail = TRUE))
+#' hist(x); abline(v = 0, col = "red", lwd = 2, lty = 2)
+#' pvalue(x) # should be 0.05 * 2
+#' pvalue(x, side = "left") # should be 0.95
+#' pvalue(x, side = "right") # should be 0.05
+pvalue <- function(x, threshold = 0, side = "both", na_rm = FALSE) {
   chk_numeric(x)
   chk_number(threshold)
+  chk_string(side)
+chk_subset(side, c("left", "right", "both"))
```
aylapear: I think this needs an indent

aylapear (approved):
I do approve but I think there is one minor formatting change that should be fixed before merging in.

R/pvalue.R
StefanoMezzini: fixed in a02744561795e4bf6be225c22f13dc46e68a7c65

StefanoMezzini:
will need to merge https://github.com/poissonconsulting/extras/pull/103 and then #102 into this branch once they are is approved

joethorley (approved):
Looks great

R/pvalue.R
```diff
@@ -7,6 +7,7 @@
 #'
 #' @param x A numeric vector of MCMC values.
 #' @inheritParams params
+#' @param side character of length 1: whether to calculate p-values for the left tail (`"left"`), right tail (`"right"`), or two-sided (`"both"`; default).
```
nehill197: Should be "A character ..." (sentence case and following the style of the rest of the package).

R/pvalue.R
```diff
@@ -15,10 +16,16 @@
 #' Academic Press, Boston. Available from <https://www.vogelwarte.ch/en/research/population-biology/book-bpa/>.
 #' @export
 #' @examples
-#' pvalue(as.numeric(0:100))
-pvalue <- function(x, threshold = 0, na_rm = FALSE) {
+#' x <- rnorm(1e6, qnorm(0.05, lower.tail = TRUE))
+#' hist(x); abline(v = 0, col = "red", lwd = 2, lty = 2)
```
nehill197: Curious the rationale for having this plot in the examples?

nehill197 (changes requested):

R/pvalue.R
StefanoMezzini: fixed in commit 231602daba7fb3ce2e05a9dfc8f6670017a5c109

R/pvalue.R
StefanoMezzini: really just to help visualize the example, since some people can get confused about the directions and the consequences of `mean = qnorm(0.05)` in `rnorm()`

nehill197:
Looks good, just didn't get an answer on my comment about having the plot in the examples?

nehill197 (approved):

aylapear (approved):
Approving assuming that the actions are green before merging in once the other PRs are merged in.

StefanoMezzini (commented):
I forgot to submit the responses 🤦

R/pvalue.R
StefanoMezzini: dropped the plot in 22ceb2216a3f2b81aee03c8af807cfdfe950c629

PR #102, opened 2026-04-24
Fix gh actions 2026 04

StefanoMezzini:


StefanoMezzini:
the failed check from `pkgdown.yml` is addressed in https://github.com/poissonconsulting/extras/pull/103

StefanoMezzini:
should merge #103 into this branch once it is approved.

.github/workflows/test-coverage.yaml
```diff
@@ -35,14 +34,16 @@ jobs:
             clean = FALSE,
             install_path = file.path(normalizePath(Sys.getenv("RUNNER_TEMP"), winslash = "/"), "package")
           )
+          print(cov)
```
nehill197, line 37: It's been forever since I worked with actions so forgive my ignorance but should the print call be in here?

StefanoMezzini: no idea 🤷 I used the file generated by `usepois::use_github_action_test_coverage()`

aylapear: We are just using the templates from the r-lib library so not sure why they have it in there. Maybe so you can see the results within the output of the action. Generally we don't change them much from the templates unless we have additional things that need to be added.

R/empty-commit-test.txt
```diff
@@ -0,0 +1 @@
+ubuntu-latest (oldrel-4)
```
aylapear: This file should be deleted. I think it was pushed by accident.

aylapear (changes requested):
Generally things look good except this text file that was added. I think we can switch to the `check-standard` template for this repo because that looks more similar to the checks we were running before. Sorry that was my bad not catching that earlier. This will fix the issue with that ubuntu with the oldest R version.

.github/workflows/test-coverage.yaml
nehill197: Good enough for me! Just thought it might have been added to check the action is working.

R/empty-commit-test.txt
StefanoMezzini: this was my test for the actions. I was planning to delete it once all tests passed, which I guess I can do now since #103 passed.

nehill197 (approved):

R/empty-commit-test.txt
StefanoMezzini: fixed in 9d1b253f960c4e4e124f8f7f789d563235c543f4

PR #105, opened 2026-04-29
allow 1-sided s-values

StefanoMezzini:
* `svalue()` can now produce one-sided s-values
* `svalue()` now has checks for all arguments (previously missing)
* `pvalue()` now only performs the sample size when the output is otherwise 0, which places a limit on how small the p-value can be that depends on the size of `x` (with cascading effects on `svalue()`
* `pvalue()` now checks that `na_rm` is logical (previously missing)
* updated descriptions of `pvalue()` and `svalue()`
* ignored posit AI files (irrelevant)
* updated docs accordingly


R/pvalue.R
```diff
@@ -9,7 +9,7 @@
 #'
 #' Note that the function contains the sample-size correction
 #' \eqn{p_{c} = p * n / (n + 1)} to avoid p-values of 0. The function can still
-#' return p-values of 1 (unlike [`probability_direction()`]).
+#' return p-values of 1.
```
joethorley: Is this when all the values are the threshold?

R/svalue.R
```diff
@@ -2,21 +2,38 @@
...
+  chk_null_or(side, vld = vld_subset, values = c("left", "right", "both"))
+  chk_number(threshold)
+  chk_logical(na_rm)
+
+  if (is.null(side)) {
+    rlang::warn("`side` should now be specified. Using `side = 'both'` by default. This warning will be removed in the future.")
+    side <- "both"
+  }
+
+  -log2(pvalue(x, side = side, threshold = threshold, na_rm = na_rm))
```
joethorley:
```suggestion
svalue <- function(x, side = "both", threshold = 0, na_rm = FALSE) {
  chk_numeric(x)
  chk_string(side)
  chk_subset(side, values = c("left", "right", "both"))
  chk_number(threshold)
  chk_logical(na_rm)

  -log2(pvalue(x, side = side, threshold = threshold, na_rm = na_rm))
```

As the previous behavior was equivalent to `side = "both"` I don't think we need to issue a warning in this case.

R/svalue.R
```diff
@@ -2,21 +2,38 @@
...
+#' in a single attempt to achieve the same probability: \eqn{-\log_2(p)}, where
+#' \eqn{p} is the p-value of interest.
 #'
 #' @param x A numeric object of MCMC values.
 #' @inheritParams params
+#' @param side A character indicating whether to calculate s-values using
+#' p-values for the left tail (`"left"`), right tail (`"right"`), or both tails
+#' (`"both"`).
+#' Defaults to `NULL`, which is treated as `"both"` but returns a warning.
+#' The warning will be removed in future package versions.
```
joethorley:
```suggestion
```

see suggested edit below

R/pvalue.R
```diff
@@ -37,22 +48,20 @@ pvalue <- function(x, threshold = 0, side = "both", na_rm = FALSE) {
     return(NA_real_)
```
joethorley:
```suggestion
    return(NA_real_)
    if(vld_true(na_rm)) {
    	x <- x[!is.na(x)]
    }
```

to ensure missing values dropped?

R/pvalue.R
```diff
@@ -37,22 +48,20 @@ pvalue <- function(x, threshold = 0, side = "both", na_rm = FALSE) {
...
   }
 
+  n <- length(x)
```
joethorley, line 49:
```suggestion
  n <- length(x)
  
  if(n == 0) {
  	 return(NA_real_)
  }
```

R/pvalue.R
```diff
@@ -37,22 +48,20 @@ pvalue <- function(x, threshold = 0, side = "both", na_rm = FALSE) {
...
-    p <- (s + 1) / (n + 1) # avoid pvalues of 0
+    s <- sum(x <= threshold) # include threshold samples
   } else if (side == "right") {
-    s <- sum(x >= threshold) # include threshold values
-    n <- length(x)
-    p <- (s + 1) / (n + 1) # avoid pvalues of 0
+    s <- sum(x >= threshold) # include threshold samples
   }
+  p <- s / n
+  p <- max(p, 1 / (n + 1)) # avoid pvalues of 0
```
joethorley, line 66: I guess my only question here is why only do the adjustment for small sample size when pvalue = 0?

joethorley (changes requested):
I think we're almost there. Just some suggested code edits to deal with edge cases, refactor a little, and a question

R/pvalue.R
StefanoMezzini: the code was written as an improper if/else:
```
  if (anyNA(x)) {
    if (vld_false(na_rm)) {
      return(NA_real_)
    }
    # continues only if ! vld_false(na_rm)
    x <- as.vector(x)
    x <- x[!is.na(x)]
  }
```

I've fixed this with commit c26115a3ee74e5f4a15e297669f4bdffdce5a46c

R/pvalue.R
StefanoMezzini: I set it up this way so that it makes the correction to place a boundary on the p-value but is otherwise unbiased. It shouldn't make a big difference if the size of `x` is large, but the limit could cause confusion with shorter vectors

R/pvalue.R
StefanoMezzini: addressed in commit 5e8f1d61d2af7a33706afd192d5b986ff4acab7b

joethorley (approved):
awesome

aylapear:
The changes in this code are also present in the other PR https://github.com/poissonconsulting/extras/pull/106 and I made comments on this code over there.

nehill197 (approved):
Good work Stefano! Looks good to me 👍

aylapear:
@StefanoMezzini Are you planning on merging this branch as well as the other one in or are you closing this one since the changes are also on the other branch? Just asking as I don't want to delay you if you are waiting for my approval on this one.

StefanoMezzini:
@aylapear the docs and tests for `svalue()` are different, so I'll still merge this into main

aylapear (approved):

Issue #98, opened 2026-04-21
Update how s-values captured


joethorley:
Use

I(p) - I(1-p)

instead of I(p*2)

StefanoMezzini:
also add a `TRUE`/`FALSE` argument that allows the current behavior but warns users that the functionality will be deprecated in the future and should be avoided

StefanoMezzini:
after the new PRs and edits, `svalue()` still returns regular s-values, but `directional_information()` now returns the directional information `log(p) - log(1-p)`

nadinehussein:
implemented

PR #106, opened 2026-04-29
add `probability_direction()` and `directional_information()`

StefanoMezzini:


R/probability-direction.R
```diff
@@ -0,0 +1,80 @@
...
+#' Makowski, D., Ben-Shachar, M.S., Chen, S.H.A., and Lüdecke, D. 2019. Indices of Effect Existence and Significance in the Bayesian Framework. Front. Psychol. 10: 2767. \doi{10.3389/fpsyg.2019.02767}.
+#' @export
+#' @examples
+#' x <- rnorm(1e6, qnorm(0.05, lower.tail = TRUE))
+#' probability_direction(x) # unspecified side returns a warning
+#' probability_direction(x, side = "left") # no warning
+#' probability_direction(x, side = "right") # = 1 - probability_direction(x, side = "left")
+#' probability_direction(c(0, 0, 1), side = "right") # does not include threshold
+#' probability_direction(c(1, 1), side = "right") # p = 1 - 1/(n+1)
+probability_direction <- function(x, side = NULL, threshold = 0, na_rm = FALSE) {
```
joethorley:
```suggestion
probability_direction <- function(x, side ="median", threshold = 0, na_rm = FALSE) {
```

I'm thinking tat rather than side = NULL and issuing a warning that we just have side = "median" by default. This will obviously necessitate some changes to the code and documentation.

R/directional-information.R
```diff
@@ -0,0 +1,125 @@
...
+  if (! length(x)) {
+    return(NA_real_)
+  }
+
+  if (is.null(side)) {
+    if (median(x = x, na.rm = na_rm) < threshold) {
+      side <- "left"
+    } else if (median(x = x, na.rm = na_rm) > threshold) {
+      side <- "right"
+    } else if (median(x, na.rm = na_rm) == threshold) {
```
joethorley:
```suggestion
    if (median(x) < threshold) {
      side <- "left"
    } else if (median(x) > threshold) {
      side <- "right"
    } else  {
```

as we have filtered out NAs upstream the na.rm argument is now redundant. Also the last else if must be true so can just use else

R/directional-information.R
```diff
@@ -0,0 +1,125 @@
...
+    p_l <- p_l + p_t / 2
+    p_r <- p_r + p_t / 2
+  } else if (threshold_split == "proportional") {
+    p_l <- p_l + p_t * (p_l / p_lr)
+    p_r <- p_r + p_t * (p_r / p_lr)
+  } else if (threshold_split == "exclude") { # effectively same as proportional
+    p_t <- 0
+    p_l <- p_l / p_lr
+    p_r <- p_r / p_lr
+  }
```
joethorley:
```suggestion
  } else { # proportional and exclude are effectively the same
    p_l <- p_l + p_t * (p_l / p_lr)
    p_r <- p_r + p_t * (p_r / p_lr)
  }
```

We can simplify the code down here

R/directional-information.R
```diff
@@ -0,0 +1,125 @@
...
+
+  if (side == "left") {
+    i <- log2(p_l) - log2(p_r)
+  } else if (side == "right") {
+    i <- log2(p_r) - log2(p_l)
+  }
```
joethorley:
```suggestion
  } else {
    i <- log2(p_r) - log2(p_l)
  }
```

simplify

R/directional-information.R
```diff
@@ -0,0 +1,125 @@
...
+
+  i <- min(i, n) # max information difference is a bit for each sample
+  i <- max(i, -n)
```
joethorley, line 103: I think this is a really a great way of dealing with it!

joethorley (changes requested):
Looks good - I like the setting bounds on the information based on the sample size.

R/probability-direction.R
StefanoMezzini: makes sense! see https://github.com/poissonconsulting/extras/issues/107

joethorley:
closes #98

R/direction.R
```diff
@@ -0,0 +1,51 @@
...
+  } else {
+    est <- estimate(x)
+    chk_number(est)
+
+    if (est < threshold) {
+      "left"
+    } else { # threshold values go with right
+      "right"
+    }
+  }
```
joethorley:
```suggestion
  }
  
  est <- estimate(x)
  chk_number(est)

 if (est < threshold) {
   return("left")
  }   
 "right"
```

The else is not needd

R/direction.R
```diff
@@ -0,0 +1,51 @@
...
+  chk_number(threshold)
+  chk_logical(na_rm)
+  chk_function(estimate)
+
+  if (na_rm) {
+    x <- x[! is.na(x)]
+  }
+
+  if (length(x) == 0) {
+    return(NA_real_)
```
joethorley:
```suggestion
    return(NA_character_)
```

In general functions should always return objects of the same type.

R/direction.R
```diff
@@ -0,0 +1,51 @@
...
+  } else {
+    est <- estimate(x)
+    chk_number(est)
```
joethorley:
```suggestion
    if(!vld_number(est)) {
    	err("The estimate function must return a number (non-missing numeric scalar).")
    }
```

Makes the error message more user friendly

R/direction.R
```diff
@@ -0,0 +1,51 @@
...
+#' calculated using a user-specified function, including the `[median]` (default),
+#' `[mean]`, geometric mean, mode, or any other custom function. By convention,
+#' values below the threshold fall to the left, while values above the threshold
+#' fall to the right. Center estimates equal to the threshold are assumed to
+#' fall to the right.
+#'
+#' @param x A numeric vector of MCMC values or any other numeric vector of samples.
+#' @param estimate A function for estimating the center of the distribution.
+#' Defaults to [`median()`], but can also be `[mean()`] or any custom function
+#' that returns a numeric vector of length 1.
```
joethorley:
```suggestion
#' that returns a number (non-missing numeric vector of length 1).
```

tweak description

R/direction.R
```diff
@@ -0,0 +1,51 @@
...
+#' `NA` values are dropped before calling the function.
+#' @param threshold A number of the threshold value.
+#' @inheritParams params
+#' @return A character vector of length one indicating if at least half of the
+#' observations are above the threshold (`"right"`) or not (`"left"`).
```
joethorley:
```suggestion
#' @return A string indicating if at least half of the
#' observations are above the threshold (`"right"`) or not (`"left"`) or NA_character_ in the case of non-missing values if `na_rm = FALSE`.
```

R/directional-information.R
```diff
@@ -0,0 +1,112 @@
+#' Directional information
+#'
+#' The information contained in a posterior belief can be summarized using the
+#' directional probability contained within that posterior.
```
joethorley:
```suggestion
#' Quantifies the information about direction in a posterior distribution based on the directional probability.
```

R/directional-information.R
```diff
@@ -0,0 +1,112 @@
...
+#' This function calculates such information using absolute difference in the the probability of direction (see [`probability_direction()`]), after converting each probability to bits (also see [`svalue()`].
+#'
+#' @param x A numeric vector of MCMC values.
+#' @param side A character vector of length 1 indicating whether to calculate
```
joethorley:
```suggestion
#' @param side A string indicating whether to calculate
```

We can define what a string is in the package if not already done.

R/directional-information.R
```diff
@@ -0,0 +1,112 @@
...
+#' directional_information(c(1, 1, -1, -1)) # both sides cancel out
+#' directional_information(rnorm(1e3, mean = 0))
+#' directional_information(rnorm(1e3, mean = 1))
+#' directional_information(rnorm(1e3, mean = 10)) # all coin flips are positive
+#' directional_information(rnorm(1e3, mean = -10)) # all coin flips are negative
+#' directional_information(rnorm(1e3, mean = 1e3)) # only quantiles matter
+#' directional_information(rnorm(1e6, mean = 1e3)) # more `x` implies more info
+
+directional_information <- function(x, side = NULL, threshold = 0,
+                                    threshold_split = "proportional",
```
joethorley, line 48: i think we should just make the default value "median" and not worry about dealing with NULL values or issuing warnings.

R/directional-information.R
```diff
@@ -0,0 +1,112 @@
...
+#' The information contained in a posterior belief can be summarized using the
+#' directional probability contained within that posterior.
+#' This function calculates such information using absolute difference in the the probability of direction (see [`probability_direction()`]), after converting each probability to bits (also see [`svalue()`].
+#'
+#' @param x A numeric vector of MCMC values.
+#' @param side A character vector of length 1 indicating whether to calculate
+#' the directional information relative to the left side (`"left"`; `x < threshold`),
+#' or the right side (`"right"`; `x > threshold`). Positive information suggests
+#' greater evidence for the specified side.
+#' Defaults to `NULL`, which uses the side of the median of `x` via [`direction()`].
```
joethorley: I think we should make the default "median" and not worry about NULL.

R/directional-information.R
```diff
@@ -0,0 +1,112 @@
...
+#' @param threshold_split A character vector of length 1 indicating how to deal
+#' with threshold values:
+#'
+#' - `"left"` to include them on the left side,
+#' - `"right"` to include them on the right side,
+#' - `"equal"` to split them equally between the left and side,
+#' - `"proportional"` (default) to split them between the left and right sides
+#'    proportionally to the values of `x` on the left and right sides,
+#' - `"exclude"` to drop the values of `x` equal to `threshold` (effectively
+#'   identical to using `"proportional"`).
```
joethorley:
```suggestion
#' - `"exclude"` to drop the values of `x` equal to `threshold` 
#'   (identical to using `"proportional"`).
```

R/directional-information.R
```diff
@@ -0,0 +1,112 @@
...
+#' directional probability contained within that posterior.
+#' This function calculates such information using absolute difference in the the probability of direction (see [`probability_direction()`]), after converting each probability to bits (also see [`svalue()`].
+#'
+#' @param x A numeric vector of MCMC values.
+#' @param side A character vector of length 1 indicating whether to calculate
+#' the directional information relative to the left side (`"left"`; `x < threshold`),
+#' or the right side (`"right"`; `x > threshold`). Positive information suggests
+#' greater evidence for the specified side.
+#' Defaults to `NULL`, which uses the side of the median of `x` via [`direction()`].
+#' @param threshold_split A character vector of length 1 indicating how to deal
```
joethorley:
```suggestion
#' @param threshold_split A string indicating how to deal
```

R/directional-information.R
```diff
@@ -0,0 +1,112 @@
...
+#'
+#' - `"left"` to include them on the left side,
+#' - `"right"` to include them on the right side,
+#' - `"equal"` to split them equally between the left and side,
+#' - `"proportional"` (default) to split them between the left and right sides
+#'    proportionally to the values of `x` on the left and right sides,
+#' - `"exclude"` to drop the values of `x` equal to `threshold` (effectively
+#'   identical to using `"proportional"`).
+#'
+#' Users should choose the combination of `side` and `threshold_split` carefully.
```
joethorley:
```suggestion
```

I think this applies to most function arguments and so doesn't need stating.

R/directional-information.R
```diff
@@ -0,0 +1,112 @@
...
+#' @inheritParams params
+#' @return A non-negative number.
```
joethorley:
```suggestion
#' @return A number indicating the directional information.
```

Now it can be negative

R/directional-information.R
```diff
@@ -0,0 +1,112 @@
...
+#' @family summary
```
joethorley, line 27:
```suggestion
#' @family summary
#' @seealso [`probability_direction()`] and [`svalue()`]
```

the see also tag is nice to use.

joethorley (changes requested):
I've added some more suggested edits.

I don't think we should be using NULL for the defaults in this case as it requires the user to read the documentation to discover what it is and adds extra complexity handling the NULL.

R/svalue.R
```diff
@@ -2,21 +2,38 @@
...
+  chk_numeric(x)
+  chk_null_or(side, vld = vld_string)
+  chk_null_or(side, vld = vld_subset, values = c("left", "right", "both"))
+  chk_number(threshold)
+  chk_logical(na_rm)
+
+  if (is.null(side)) {
+    rlang::warn("`side` should now be specified. Using `side = 'both'` by default. This warning will be removed in the future.")
+    side <- "both"
+  }
```
joethorley:
```suggestion
svalue <- function(x, side = "both", threshold = 0, na_rm = FALSE) {
  chk_numeric(x)
  chk_string(side)
  chk_subset(side, values = c("left", "right", "both"))
  chk_number(threshold)
  chk_logical(na_rm)
```

joethorley (changes requested):
use "both" instead of NULL in svalue

R/direction.R
StefanoMezzini: sorry, that was a typo

R/direction.R
StefanoMezzini: definitely. addressed in commit 0506ce558ba05c13625a030c39734a0a12b1df6b since the edit was outdated

R/directional-information.R
StefanoMezzini: already present in `params.R`: https://github.com/poissonconsulting/extras/blob/8c0878ee3b8c65ae2413a7a6cf81b51656cfc39b/R/params.R#L8

I just saw the script this morning

R/directional-information.R
StefanoMezzini: fixed in commit ad533c1a9ab0eac5c7473a7f0204da5595fe9818

R/directional-information.R
StefanoMezzini: fixed in commit ad533c1a9ab0eac5c7473a7f0204da5595fe9818

R/directional-information.R
StefanoMezzini: good to know. thanks! I'll update it in all other necessary functions

StefanoMezzini: reverted the commit (48c4f6be7698c71cd1e998de2c1fa25d3e7976d3) since the functions are in the same family

R/svalue.R
StefanoMezzini: I had made the changes on the `1sided-svalue` branch. I've merged the changes in commit ba2390441f55d2c83455306b7418361c9fefee42

R/directional-information.R
```diff
@@ -0,0 +1,116 @@
...
+  if (anyNA(x)) {
+    if (vld_true(na_rm)) {
+      x <- as.vector(x)
+      x <- x[!is.na(x)]
+    } else {
+      return(NA_real_)
+    }
+  }
+
+  if (any(is.infinite(x))) {
```
joethorley: Aren't finite values still just positive (or negative) like any other value?

R/probability-direction.R
```diff
@@ -0,0 +1,80 @@
...
+  if (anyNA(x)) {
+    if (vld_true(na_rm)) {
+      x <- as.vector(x)
+      x <- x[!is.na(x)]
+    } else {
+      return(NA_real_)
+    }
+  }
+
+  if (any(is.infinite(x))) {
```
joethorley: Can't infinite values be treated like any other value - they are just guaranteed to be positive or negative?

R/probability-direction.R
joethorley: Don't you need to change tho - I'm still seeing code reference NULL in the function?

R/probability-direction.R
```diff
@@ -0,0 +1,80 @@
...
+    return(NA_real_)
+  }
+
+  n <- length(x)
+  if (n == 0) {
+    return(NA_real_)
+  }
+
+  if (is.null(side)) {
```
joethorley:
```suggestion
  if (side == "median") {
```

see comment about dropping NULL

R/direction.R
```diff
@@ -0,0 +1,57 @@
...
+#' direction(c(-100, 1, 1), mean)
+#' direction(c(100, 0.01, 0.01), function(.x) exp(mean(log(.x))))
+direction <- function(x, estimate = median, threshold = 0, na_rm = FALSE) {
+  chk_numeric(x)
+  chk_number(threshold)
+  chk_flag(na_rm)
+  chk_function(estimate)
+
+  if (anyNA(x)) {
+    if (vld_true(na_rm)) {
```
joethorley:
```suggestion
    if (na_rm) {
```

As we have checkd if flag we don't need `vld_true()` in this instance.

R/svalue.R
```diff
@@ -2,21 +2,31 @@
...
-svalue <- function(x, threshold = 0, na_rm = FALSE) {
-  -log(pvalue(x, threshold = threshold, na_rm = na_rm), 2)
+#' svalue(rnorm(1e4, mean = 1), side = "left")
+#' svalue(rnorm(1e4, mean = 1), side = "right")
+svalue <- function(x, side = "both", threshold = 0, na_rm = FALSE) {
+  chk_numeric(x)
+  chk_string(side)
+  chk_subset(side, values = c("left", "right", "both"))
+  chk_number(threshold)
+  chk_logical(na_rm)
```
joethorley:
```suggestion
```

As all the same checks are in pvalue and the arguments have the same name we let pvalue do all the argument testing for both functions.

joethorley (changes requested):
Some more tweaks

R/probability-direction.R
StefanoMezzini: missed that!

R/directional-information.R
StefanoMezzini: fixed in commit 2eebcc4f7c4d2e66e894ffeba1bee16e18d233c4

R/probability-direction.R
StefanoMezzini: fixed in commit 2eebcc4f7c4d2e66e894ffeba1bee16e18d233c4

R/directional-information.R
StefanoMezzini: I think I had it in other functions in case we used the mean for measuring the center, but this makes sense. Even if one uses the mean, the output should still be appropriate

R/direction.R
```diff
@@ -0,0 +1,57 @@
...
+#' direction(c(100, 0.01, 0.01), function(.x) exp(mean(log(.x))))
+direction <- function(x, estimate = xtr_median, threshold = 0, na_rm = FALSE) {
+  chk_numeric(x)
+  chk_number(threshold)
+  chk_flag(na_rm)
+  chk_function(estimate)
+
+  if (anyNA(x)) {
+    if (na_rm) {
+      x <- as.vector(x)
```
joethorley, line 38:
```suggestion
x <- as.vector(x)
  if (anyNA(x)) {
    if (na_rm) {
```

I think we should convert to a vector so length(x) is valid if no missing values.

R/directional-information.R
```diff
@@ -0,0 +1,112 @@
...
+                                    na_rm = FALSE) {
+  chk_numeric(x)
+  chk_subset(side, c("left", "right", "median"))
+  chk_number(threshold)
+  chk_subset(threshold_split, c("left", "right", "equal", "proportional", "exclude"))
+  chk_flag(na_rm)
+
+  if (anyNA(x)) {
+    if (na_rm) {
+      x <- as.vector(x)
```
joethorley, line 58:
```suggestion
x <- as.vector(x)
  if (anyNA(x)) {
    if (na_rm) {
```

ensure vector for later length(x) test

joethorley (approved):
Looks good - just two minor related changes to ensure a vector even if no missing values for when use length(x)

R/direction.R
StefanoMezzini: agreed that it makes more sense to have it earlier -- this was code I copied from `pvalue()`, I believe. Is `x` not already guaranteed to be a vector based on the checks though?

StefanoMezzini: not necessary -- it only needs to be a vector if we use `[` for dropping `NA`s

R/directional-information.R
StefanoMezzini: not necessary -- it only needs to be a vector if we use `[` for dropping `NA`s

R/directional-information.R
```diff
@@ -0,0 +1,112 @@
...
+  i <- min(i, n) # max information difference is a bit for each sample
+  i <- max(i, -n)
+  # equivalent to, if o is the odds ratio:
+  # if (is.infinite(o)) {
+  #   if (o > 0) {
+  #     o <- n # = n / (n+1) / (1 / (n+1))
+  #   } else {
+  #     o <- 1 / n # = 1 / (n+1) / (n / (n+1))
+  #   }
+  # }
```
aylapear: If this isn't being used it should be removed or a comment of why it is commented out.

R/probability-direction.R
```diff
@@ -0,0 +1,76 @@
...
+#' distribution above or below zero.
+#' By default, the direction is based on the side of the median value, but it
+#' can be specified to measure support for specific hypotheses.
+#' A right-side PD of 0.9 indicates that the CRI spanning from the threshold to
+#' infinity has a coverage of 90%.
+#' Can be used as a measure of certainty in the direction of the estimate
+#' (e.g., positive or negative when using a threshold of 0).
+#' **NOTE:** probability estimates of 0 or 1 are corrected towards 0.5 by adding
+#' or subtracting `1 / (length(x) + 1)`, where `x` is a vector of MCMC samples.
+#' Ideally, `x` should be large enough as to make the correction negligible.
```
aylapear: The description section should say what the function does in one or two sentences. A bunch of this should be moved to the details section where you can provide additional info on the how, why, edge cases, use cases, etc.

R/probability-direction.R
```diff
@@ -0,0 +1,76 @@
...
+#' @return A number between 0 and 1.
+#' @family summary
+#' @references
+#' Makowski, D., Ben-Shachar, M.S., Chen, S.H.A., and Lüdecke, D. 2019. Indices of Effect Existence and Significance in the Bayesian Framework. Front. Psychol. 10: 2767. \doi{10.3389/fpsyg.2019.02767}.
+#' @export
+#' @examples
+#' x <- rnorm(1e6, qnorm(0.05, lower.tail = TRUE))
+#' probability_direction(x, side = "left")
+#' probability_direction(x, side = "right") # = 1 - probability_direction(x, side = "left")
+#' probability_direction(c(0, 0, 1), side = "right") # does not include threshold
```
aylapear: I don't understand the `#does not include threshold` comment. I don't see threshold provided beyond the default in any of the examples.

R/probability-direction.R
```diff
@@ -0,0 +1,76 @@
...
+#' probability_direction(c(1, 1), side = "right") # p = 1 - 1/(n+1)
+probability_direction <- function(x, side = NULL, threshold = 0, na_rm = FALSE) {
+  chk_numeric(x)
+  chk_null_or(side, vld = vld_string)
+  chk_null_or(side, vld = vld_subset, values = c("left", "right"))
```
aylapear: I believe the `chk_null_or(side, vld = vld_subset, values = c("left", "right"))` will take care of the type so line 37 can be removed.

R/pvalue.R
```diff
@@ -20,39 +35,39 @@
 #' pvalue(x) # should be 0.05 * 2
 #' pvalue(x, side = "left") # should be 0.95
 #' pvalue(x, side = "right") # should be 0.05
-pvalue <- function(x, threshold = 0, side = "both", na_rm = FALSE) {
+pvalue <- function(x, side = "both", threshold = 0, na_rm = FALSE) {
```
aylapear, line 39: Changing the argument order should only be done if really needed and it should be clearly noted in the NEWS file.

R/svalue.R
```diff
@@ -2,21 +2,26 @@
...
 #' The American Statistician 73(sup1): 106–114.
 #' \doi{10.1080/00031305.2018.1529625}.
 #' @export
 #' @examples
-#' svalue(as.numeric(0:100))
-svalue <- function(x, threshold = 0, na_rm = FALSE) {
-  -log(pvalue(x, threshold = threshold, na_rm = na_rm), 2)
+#' svalue(rnorm(1e4, mean = 1), side = "left")
+#' svalue(rnorm(1e4, mean = 1), side = "right")
+svalue <- function(x, side = "both", threshold = 0, na_rm = FALSE) {
```
aylapear, line 28: Changing the argument order should only be done if really needed and it should be clearly noted in the NEWS file.

R/svalue.R
```diff
@@ -2,21 +2,26 @@
...
+#' Greenland, S. 2019. Valid P-Values Behave Exactly as They Should:
+#' Some Misleading Criticisms of P-Values and Their Resolution With S -Values.
 #' The American Statistician 73(sup1): 106–114.
 #' \doi{10.1080/00031305.2018.1529625}.
 #' @export
 #' @examples
-#' svalue(as.numeric(0:100))
-svalue <- function(x, threshold = 0, na_rm = FALSE) {
-  -log(pvalue(x, threshold = threshold, na_rm = na_rm), 2)
+#' svalue(rnorm(1e4, mean = 1), side = "left")
```
aylapear, line 26: I'd keep the old example in unless it is wrong as it shows a simpler use of the function.

tests/testthat/test-direction.R
```diff
@@ -0,0 +1,16 @@
+test_that("`direction()` returns the correct direction.", {
```
aylapear, line 1: If you run `covr::file_report(covr::file_coverage("R/direction.R", "tests/testthat/test-direction.R"))` you can see which lines of code are still not being tested. Each line of the function should have at least 1 test for it. Looks like you need to add some for the `na_rm` argument and to trigger the `err` error.

tests/testthat/test-directional-information.R
```diff
@@ -0,0 +1,45 @@
+test_that("directional_information() returns correct values.", {
```
aylapear, line 1: There are still some lines of the function that are not being tested like the `threshold_split` argument.

R/probability-direction.R
```diff
@@ -0,0 +1,76 @@
...
+  }
+
+  n <- length(x)
+  if (n == 0) {
+    return(NA_real_)
+  }
+
+  if (side == "median") {
+    side <- direction(x)
+  }
```
aylapear, line 61: `side` needs to be updated or this code removed as it can't become the value `media`

aylapear (changes requested):

R/probability-direction.R
joethorley: It doesn't do any enforcement on length vld_string makes sure its length 1 so we should keep

aylapear: That is good to know. I will watch for that in my own code.

R/directional-information.R
StefanoMezzini: that's why I included
```
# equivalent to, if o is the odds ratio:
```
I kept it in there to help frame the decisions around the cutoff

R/probability-direction.R
StefanoMezzini: the threshold (0) is excluded from the probability, meaning that it returns `P(X >0) = 1/3` instead of `P(X >= 0) = 1`

R/probability-direction.R
StefanoMezzini: fixed in commit 0e54068b9dfa27e100b8afdd4212e846e2591afa

R/pvalue.R
StefanoMezzini: this was a request from @joethorley, since the `side` will likely be changed more often than the `threshold`. I think the impact of changing the order will be minimal since the `side` argument was just added in the last version. I will make a note of adding it to the NEWS explicitly though.

R/svalue.R
StefanoMezzini: fixed in 9a6b12fe46bf162f4785e85bc9ce3e60e69dc951

tests/testthat/test-direction.R
StefanoMezzini: fixed in 97e5d011bb4f49ce81298b46142e69f8617b2e3f

R/probability-direction.R
StefanoMezzini: fixed in eb63ca7f80e75498d36b35c4097c2e5fd8ee4214

R/svalue.R
StefanoMezzini: same as above

tests/testthat/test-directional-information.R
StefanoMezzini: fixed in 6197a4fbe4b1428c99a1ba5486596151610f5a25

R/probability-direction.R
aylapear: That makes more sense. I think updating the comment to not use the word "threshold" would be a benefit if possible because my initial thought was you were taking about not including the `threshold` argument which I found confusing since it has a default argument that is the same in all the examples. 


```suggestion
#' probability_direction(c(0, 0, 1), side = "right") # returns P(X >0) = 1/3 instead of P(X >= 0) = 1
```

R/directional-information.R
aylapear: That needs to be added to the comment. To me it looks like old code that used to be used or was trying something that wasn't needed and was missed being deleted. Saying it was left intentionally because of the cutoff decision would be helpful to provide more context.

aylapear (changes requested):
Two minor things to double check on.

R/probability-direction.R
StefanoMezzini: agreed. I've committed the change in commit 8fe1ae849b25715fa334b1d7558b09c8cd4898b2

R/directional-information.R
StefanoMezzini: addressed in commit edf3a0a76fcf3f9b5c443f3781994df04f619279

R/direction.R
```diff
@@ -0,0 +1,57 @@
...
+#' mean, geometric mean, mode, or any other custom function. By convention,
+#' values below the threshold fall to the left, while values above the threshold
+#' fall to the right. Center estimates equal to the threshold are assumed to
+#' fall to the right.
+#'
+#' @param x A numeric vector of MCMC values or any other numeric vector of samples.
+#' @param estimate A function for estimating the center of the distribution.
+#' Defaults to [xtr_median()], but can also be [xtr_mean()] or any custom function
+#' that returns a number (a non-missing numeric vector of length 1).
+#' `NA` values are dropped before calling the function.
```
nehill197: Isn't the NA behaviour controlled by the `na_rm` argument?

R/direction.R
```diff
@@ -0,0 +1,57 @@
...
+#' @param threshold A number of the threshold value.
+#' @inheritParams params
+#' @return A string indicating if at least half of the
+#' observations are above the threshold (`"right"`) or not (`"left"`) or NA_character_ in the case of non-missing values if `na_rm = FALSE`.
```
nehill197: NA_character_ in the case of *non-missing* values? 
Wouldn't it be in the case of missing values?

R/direction.R
```diff
@@ -0,0 +1,57 @@
...
+      return(NA_character_)
+    }
+  }
+
+  if (length(x) == 0) {
+    return(NA_character_)
+  }
+
+  est <- estimate(x)
+  if(!vld_number(est)) {
```
nehill197: nit-picky of me but could you add a space after the `if` to be consistent?

R/probability-direction.R
```diff
@@ -0,0 +1,78 @@
...
+#' Defaults to `"median"`, which uses the side of the median of `x` via
+#' [`direction()`].
+#' @param threshold A number of the threshold value, which is excluded from the
+#' interval for the probability.
+#' @inheritParams params
+#' @return A number between 0 and 1.
+#' @details
+#' By default, the direction is based on the side of the median value, but it
+#' can be specified to measure support for specific hypotheses.
+#' A right-side PD of 0.9 indicates that the CRI spanning from the threshold to
```
nehill197: CRI not defined?

R/directional-information.R
```diff
@@ -0,0 +1,108 @@
+#' Directional information
+#'
+#' Quantifies the information about direction in a posterior distribution based on the directional probability.
+#' This function calculates such information using absolute difference in the the probability of direction (see [`probability_direction()`]), after converting each probability to bits (also see [`svalue()`].
```
nehill197: using "the" absolute difference?

R/directional-information.R
```diff
@@ -0,0 +1,108 @@
...
+#' @param threshold_split A string indicating how to deal
+#' with threshold values:
+#'
+#' - `"left"` to include them on the left side,
+#' - `"right"` to include them on the right side,
+#' - `"equal"` to split them equally between the left and side,
+#' - `"proportional"` (default) to split them between the left and right sides
+#'    proportionally to the values of `x` on the left and right sides,
+#' - `"exclude"` to drop the values of `x` equal to `threshold`
+#'   (identical to using `"proportional"`).
```
nehill197, line 22: Why we have "exclude" if it's identical to using "proportional"? (i.e., why not just have "proportional"?)

R/pvalue.R
```diff
@@ -20,39 +35,39 @@
...
+  chk_number(threshold)
+  chk_flag(na_rm)
 
   if (anyNA(x)) {
-    if (vld_false(na_rm)) {
+    if (na_rm) {
+      x <- as.vector(x)
+      x <- x[!is.na(x)]
+    } else {
       return(NA_real_)
```
nehill197, line 51: Should we be explicit in the docs about it returning NA if na_rm = FALSE and anyNA(x)? (and consider this for some of the other functions in here too)

nehill197 (changes requested):
Mostly some suggestions/questions about the docs here, good work!

R/direction.R
StefanoMezzini: does my edit in e366ced4d09bba1f7ca0d37d4056395ed016d280 clarify this?

R/direction.R
StefanoMezzini: yeah, sorry! how's this? 1dcd7f4711103a8b18974967bae5585fc497cebc

R/direction.R
nehill197: yes! but you have na.rm there instead of na_rm

R/direction.R
StefanoMezzini: fixed in commit 1dcd7f4711103a8b18974967bae5585fc497cebc

R/probability-direction.R
StefanoMezzini: changed to the more generic "interval" in 069a5f47c1b486453ee8dae0f1019c0b9e5d9854

R/directional-information.R
StefanoMezzini: fixed in commit 069a5f47c1b486453ee8dae0f1019c0b9e5d9854. We also don't use the *absolute* difference because we use the sign to indicate evidence for or against

R/directional-information.R
StefanoMezzini: because most people would probably not realize that excluding the threshold is equivalent to including it proportionally to the two sides. I'm happy to drop `"exclude"` and change the docs to clarify that using proportional is equivalent to excluding them (or vice-versa).

R/pvalue.R
StefanoMezzini: fixed in af2fe46

R/direction.R
```diff
@@ -12,11 +12,12 @@
 #' @param estimate A function for estimating the center of the distribution.
 #' Defaults to [xtr_median()], but can also be [xtr_mean()] or any custom function
 #' that returns a number (a non-missing numeric vector of length 1).
-#' `NA` values are dropped before calling the function.
+#' If `na.rm` is true, `NA` values are dropped before calling the function.
```
nehill197: Should be `na_rm`!

R/pvalue.R
```diff
@@ -24,6 +24,7 @@
 #' p-values for the left tail (`"left"`), right tail (`"right"`), or two-sided (`"both"`; default).
 #' @inheritParams params
 #' @return A number between 0 and 1.
+#' If `x` has `NA` values but `na.rm` is `FALSE`, returns `NA_real`.
```
nehill197: `na_rm` here too!

R/svalue.R
```diff
@@ -11,6 +11,7 @@
 #' p-values for the left tail (`"left"`), right tail (`"right"`), or both tails
 #' (`"both"`; default).
 #' @return A non-negative number.
+#' If `x` has `NA` values but `na.rm` is `FALSE`, returns `NA_real`.
```
nehill197: And here 😁

nehill197 (changes requested):
Just these `na_rm` changes to fix!

R/svalue.R
StefanoMezzini: fixed in b4e8d51ee5a46d8467486725a1342f90e6151c65

StefanoMezzini:
fixed all five in b4e8d51ee5a46d8467486725a1342f90e6151c65

nehill197 (dismissed):

StefanoMezzini:
I'm not sure what the stale review dismissal is about, as I just merged the conflicts. This may be related: https://github.com/isaacs/github/issues/1157.

R/probability-direction.R
StefanoMezzini: this has been fixed in a previous commit (eb63ca7f80e75498d36b35c4097c2e5fd8ee4214)

nehill197 (approved):

PR #110, opened 2026-05-05
Add functions to calculate ROPE and CIs

StefanoMezzini:
The new functions calculate the Region of Practical Equivalence (ROPE) based on the (credible) intervals produced by `xtr_ci()`, which can produce three types of intervals:

- highest density intervals (HDI): cover the areas with highest density to result in the narrowest intervals with the specified coverage;
- equal tailed intervals (ETI): have the same density in the left and right tail
- normal approximation intervals: quick to calculate, but can have bad coverage if the sampled distribution is very different from normal. still, can be useful for testing and comparing to the HDI and ETI options.


StefanoMezzini:
I can add other reviewers after the first round of reviews is done.

R/ci_eti.R
```diff
@@ -0,0 +1,44 @@
...
+
+  if(length(x) == 0) {
+    return(data.frame(lower = NA_real_, upper = NA_real_))
+  }
+
+  if (anyNA(x)) {
+    if (vld_true(na_rm)) {
+      x <- x[!is.na(x)]
+    } else {
+      return(data.frame(lower = NA_real_, upper = NA_real_))
```
joethorley: I think we should return `tibble::tibble()` instead of `data.frame()` for all these functions as better behaviour and extras is already importing the tibble package.

R/ci_eti.R
```diff
@@ -0,0 +1,44 @@
...
+#' @return A data frame of the `lower` and `upper` limits for the credible interval.
+#' Note that the interval is not guaranteed to be one-sided or two-sided.
+#' @export
+#' @seealso [extras::xtr_ci], [extras::xtr_ci_hdi], and [extras::xtr_ci_norm]
+#' @examples
+#' xtr_ci_eti(rnorm(1e4))
+#' @name xtr_ci_eti
+NULL
+
+xtr_ci_eti <- function(x, level = 0.95, na_rm = FALSE, quiet = TRUE) {
```
joethorley:
```suggestion
xtr_ci_eti <- function(x, level = 0.95, ..., na_rm = FALSE) {
chk_unused(...)
```

this is a cool trick to require naming of arguments after level to ensure code is more robust. Also in this case I think we can drop the quiet argument as not being used.

joethorley: Also use ... in function below.

R/ci_hdi.R
```diff
@@ -0,0 +1,63 @@
...
+    if (vld_true(na_rm)) {
+      x <- x[!is.na(x)]
+    } else {
+      return(data.frame(lower = NA_real_, upper = NA_real_))
+    }
+  }
+  x <- sort(x)
+  n <- length(x)
+
+  if (n < 1 / (1 - level)) {
```
joethorley: I might be being dense but I'm not sure what the `/ (1 - level)` is doing in this condition?

R/ci_eti.R
joethorley: and drop quiet

R/ci_hdi.R
```diff
@@ -0,0 +1,63 @@
...
+    return(data.frame(lower = NA_real_, upper = NA_real_))
+  }
+
+  n_in <- ceiling(n * level)
+  n_out <- n - n_in
+
+  if (sum(is.infinite(x)) >= n_in) {
+    return(data.frame(lower = Inf * sign(min(x[is.infinite(x)])),
+                      upper = Inf * sign(max(x[is.infinite(x)]))))
```
joethorley:
```suggestion
    return(data.frame(lower = min(x),
                                  upper = max(x)
```
If all Inf are positive we could end up with an upper that is greater than the median so maybe we should use just the min and max?

R/ci_norm.R
```diff
@@ -0,0 +1,49 @@
...
+#' @return A data frame of the `lower` and `upper` limits for the credible interval.
+#' Note that the interval is not guaranteed to be one-sided or two-sided.
+#' @export
+#' @seealso [extras::xtr_ci], [extras::xtr_ci_eti], and [extras::xtr_ci_hdi]
+#' @examples
+#' xtr_ci_norm(rnorm(1e4))
+#' @name xtr_ci_norm
+NULL
+
+xtr_ci_norm <- function(x, level = 0.95, na_rm = FALSE, quiet = TRUE) {
```
joethorley: This function is logical but I'm not sure what a use case would be?

R/rope.R
```diff
@@ -0,0 +1,49 @@
...
+#' @export
+#' @seealso [extras::xtr_ci]
+#' @examples
+#' rope(c(-Inf, -1, -0.1, 0.1, 1, Inf))
+#' rope(rnorm(1e4))
+#' @name rope
+NULL
+
+rope <- function(x, threshold = 0, interval = c(-0.1, 0.1), ci_level = 1,
+                 type = "HDI", na_rm = FALSE, quiet = TRUE) {
```
joethorley:
```suggestion
rope <- function(x, threshold = 0, interval = c(-0.1, 0.1), ..., level = 1,
                 type = "HDI", na_rm = FALSE) {
  chk_unused(...)
```

use ... trick to enforce name matching
rename ci_level to level so consistent with other functions.
drop quiet as not being used.

R/rope.R
```diff
@@ -0,0 +1,49 @@
...
+  chk_numeric(x)
+  chk_numeric(interval)
+  chk_length(interval, 2)
+  chk_number(ci_level)
+  chk_true(all(ci_level > 0 & ci_level <= 1))
```
joethorley:
```suggestion
  chk_range(ci_level)
```

check's if between 0 and 1

joethorley (changes requested):
Its coming together into some really useful functions - see suggestions for improvements.

R/ci_eti.R
StefanoMezzini: fixed in commit 5665a3165ad06bea85a1565813d23c1a5079b00d

R/ci_hdi.R
StefanoMezzini: it helps ensure we have a large enough sample. To be able to estimate the CI, we need at least one sample outside the CI, with means we can't calculate the CI if `n * (1 - level) < 1`. For example, a 95% CI requires at least 1 observation outside the CI, which requires `n >= 1 / 0.05` to allow 19 to be in and 1 to be out.

R/ci_norm.R
StefanoMezzini: I was thinking it would be helpful for quick tests, but I agree with what we said about it being more of a cost and a nuisance than a benefit. Removed in commit e80735a425fc40e6c4e015f5f6c920c56b62077f.

R/ci_eti.R
StefanoMezzini: fixed in 08a93d0028566a53c81a41672397e49eaa1d070a

StefanoMezzini:
the checks are failing because of errors detailed in this issue https://github.com/issues/created?issue=poissonconsulting%7Cinfrastructure-issues%7C227

R/ci_hdi.R
StefanoMezzini: fixed in b5d1adbc5313ce238eeb5323bf992e2a99204054

R/rope.R
StefanoMezzini: fixed in 4dbac14297a8ca566d1673ec9cbf66dbb6882d48

R/rope.R
StefanoMezzini: fixed in 4dbac14297a8ca566d1673ec9cbf66dbb6882d48

R/ci_hdi.R
joethorley: OK i was forgetting about precedence....

but I think we should just return the max and min if all MCMC samples effectively in the CI. So that gives lower and upper bounds when few MCMC sample even tho unreliable for running code.

Note I think we should have CIs with 2 or more values.

R/ci.R
```diff
@@ -0,0 +1,34 @@
...
+  chk_subset(type, c("HDI", "ETI"))
+  chk_range(level)
+  chk_flag(na_rm)
+  chk_unused(...)
+
+  if(type == "HDI") {
+    xtr_ci_hdi(x = x, level = level, na_rm = na_rm)
+  } else if (type == "ETI") {
+    xtr_ci_eti(x = x, level = level, na_rm = na_rm)
+  }
```
joethorley:
```suggestion
    return(xtr_ci_hdi(x = x, level = level, na_rm = na_rm))
  } 
  xtr_ci_eti(x = x, level = level, na_rm = na_rm)
```

arguably slightly easier to read

R/ci_eti.R
```diff
@@ -0,0 +1,54 @@
...
+#' @name xtr_ci_eti
+NULL
+
+xtr_ci_eti <- function(x, level = 0.95, ..., na_rm = FALSE) {
+  chk_numeric(x)
+  chk_number(level)
+  chk_range(level, inclusive = TRUE)
+  chk_gt(level)
+  chk_flag(na_rm)
+  chk_unused(...)
```
joethorley:
```suggestion
  chk_unused(...)
  chk_flag(na_rm)
```

i tend to order same as arguments for maintainability

R/ci_eti.R
```diff
@@ -0,0 +1,54 @@
...
+
+  if(is.integer(x)) {
+    na <- NA_integer_
+  } else {
+    na <- NA_real_
+  }
```
joethorley:
```suggestion
na <-  if(is.integer(x)) NA_integer_ else NA_real_
```

this should also work and arguably easier to read

R/ci_eti.R
```diff
@@ -0,0 +1,54 @@
...
+
+  if(length(x) <= 1) {
+    return(data.frame(lower = na, upper = na))
+  }
+
+  if (anyNA(x)) {
+    if (vld_true(na_rm)) {
```
joethorley:
```suggestion
    if (na_rm) {
```

as checked flag no need for vld_true

R/rope.R
```diff
@@ -0,0 +1,49 @@
...
+#' credible interval) within the region of practical equivalence.
+#' @export
+#' @seealso [extras::xtr_ci]
+#' @examples
+#' rope(c(-Inf, -1, -0.1, 0.1, 1, Inf))
+#' rope(rnorm(1e4))
+#' @name rope
+NULL
+
+rope <- function(x, threshold = 0, interval = c(-0.1, 0.1), ..., level = 1,
```
joethorley:
```suggestion
xtr_rope <- function(x, threshold = 0, interval = c(-0.1, 0.1), ..., level = 1,
```

i think rename to xtr_rope.

R/rope.R
```diff
@@ -0,0 +1,49 @@
...
+                 type = "HDI", na_rm = FALSE) {
+  chk_numeric(x)
+  chk_number(threshold)
+  chk_numeric(interval)
+  chk_length(interval, 2)
+  chk_number(level)
```
joethorley, line 42:
```suggestion
  chk_unused(...)
  chk_number(level)
```

makes sure ... unused

R/rope.R
```diff
@@ -0,0 +1,49 @@
...
+#' rope(rnorm(1e4))
+#' @name rope
+NULL
+
+rope <- function(x, threshold = 0, interval = c(-0.1, 0.1), ..., level = 1,
+                 type = "HDI", na_rm = FALSE) {
+  chk_numeric(x)
+  chk_number(threshold)
+  chk_numeric(interval)
+  chk_length(interval, 2)
```
joethorley, line 38:
```suggestion
  chk_length(interval, 2)
  chk_not_any_na(interval)
  chk_sorted(interval)
```

more checks on interval

joethorley (changes requested):
Some super minor refactoring of code in xtr_ci_eti() that could also be applied to xtr_ci_hdi() and also some comments on rope().

R/ci_eti.R
StefanoMezzini: yeah, I missed that

R/ci_eti.R
StefanoMezzini: I thought our formatting style preferred using `{ }`. I prefer it shorter like that, too

R/rope.R
StefanoMezzini: fixed in ae8390b702f2248ef4b13d07bde17fc291dd30d8

R/ci_hdi.R
StefanoMezzini: fixed in 4417d6346a0c46242f0d602a514df4ab56183299

R/rope.R
```diff
@@ -0,0 +1,52 @@
...
+  if (anyNA(x)) {
+    if (vld_true(na_rm)) {
+      x <- x[!is.na(x)]
+    } else {
+      NA_real_
+    }
+  }
+
+  if(level < 1) {
+    ci <- xtr_ci(x = x, level = level, ..., type = type, na_rm = na_rm)
```
joethorley, line 60: what happens if xtr_ci returns missing values? Do we need test for this possibility by checking for x with length >= 2 and returning NA_real_ before even calling xtr_ci?

joethorley (changes requested):
question around edge case

R/rope.R
StefanoMezzini: isn't that covered by the precedent code?
```
if (anyNA(x)) {
    if (vld_true(na_rm)) {
      x <- x[!is.na(x)]
    } else {
      NA_real_
    }
  }
  ```

Comment (joethorley) [reply R/rope.R:60]: what if na_rm is true and only 1 value is left? won't HDI return missing values?

Comment (StefanoMezzini) [reply R/rope.R:60]: ah, that's a valid point:
```
> xtr_rope(c(NA_integer_, 1), na_rm = TRUE)
[1] 0
> xtr_rope(c(NA_integer_, 0), na_rm = TRUE)
[1] 1
> xtr_rope(c(NA_integer_, 0:1), na_rm = TRUE)
[1] 0.5
```

Comment (StefanoMezzini) [reply R/rope.R:60]: fixed with commit eac56055bc93f6266249b3a2a41a3daaa5f2e430

File: R/ci.R [hunk 16]
```diff
@@ -0,0 +1,33 @@
...
+#' @param na_rm A flag indicating whether to remove missing values.
+#' @return A data frame of the `lower` and `upper` limits for the credible interval.
+#' Note that the interval is not guaranteed to be one-sided or two-sided.
+#' @export
+#' @seealso [mcmcr::coef]
+#' @examples
+#' xtr_ci(rnorm(1e4), type = "HDI")
+#' @name xtr_ci
+NULL
+
```
joethorley:
```suggestion
```

these are not needed for function definitions

R/ci_eti.R
```diff
@@ -0,0 +1,50 @@
...
+#' interval.
+#' Note that the interval is guaranteed to be two-sided with real (i.e., double)
+#' numeric limits, even if the input data are integers.
+#' @export
+#' @seealso [extras::xtr_ci()] and [extras::xtr_ci_hdi()]
+#' @examples
+#' xtr_ci_eti(rnorm(1e4))
+#' @name xtr_ci_eti
+NULL
+
```
joethorley:
```suggestion
```

these are not needed for function definitions

R/ci_hdi.R
```diff
@@ -0,0 +1,92 @@
...
+#' @export
+#' @seealso [extras::xtr_ci()] and [extras::xtr_ci_eti()]
+#' @examples
+#' xtr_ci_hdi(1:10, level = 0.1) # only 10% of values inside
+#' xtr_ci_hdi(1:10, level = 0.2) # only 20% of values inside
+#' xtr_ci_hdi(1:10, level = 0.2 + 0.01) # at least 20.1% of values inside
+#' xtr_ci_hdi(1:100) # inclusive interval [3, 98] with 5% of values outside
+#' @name xtr_ci_hdi
+NULL
+
```
joethorley:
```suggestion
```

these are not needed for function definitions

R/ci.R
```diff
@@ -0,0 +1,33 @@
...
+xtr_ci <- function(x, level = 0.95, ..., type = "HDI", na_rm = FALSE) {
+  chk_numeric(x)
+  chk_number(level)
+  chk_subset(type, c("HDI", "ETI"))
+  chk_range(level)
+  chk_flag(na_rm)
```
joethorley:
```suggestion
```
let downstream functions test

R/ci.R
```diff
@@ -0,0 +1,33 @@
...
+#' @export
+#' @seealso [mcmcr::coef]
+#' @examples
+#' xtr_ci(rnorm(1e4), type = "HDI")
+#' @name xtr_ci
+NULL
+
+xtr_ci <- function(x, level = 0.95, ..., type = "HDI", na_rm = FALSE) {
+  chk_numeric(x)
+  chk_number(level)
```
joethorley:
```suggestion
```

let downstream functions test

R/ci_hdi.R
```diff
@@ -0,0 +1,92 @@
...
+xtr_ci_hdi <- function(x, level = 0.95, ..., na_rm = FALSE) {
+  chk_numeric(x)
+  chk_number(level)
+  chk_range(level, inclusive = TRUE)
+  chk_gt(level)
+  chk_flag(na_rm)
+  chk_unused(...)
```
joethorley:
```suggestion
  chk_unused(...)
  chk_flag(na_rm)
```

order by appearance in arguments

R/ci_hdi.R
```diff
@@ -0,0 +1,92 @@
...
+
+  if(is.integer(x)) {
+    na <- NA_integer_
+  } else {
+    na <- NA_real_
+  }
```
joethorley:
```suggestion
na <- if(is.integer(x)) NA_integer_ else NA_real_
```

optional

R/ci_hdi.R
```diff
@@ -0,0 +1,92 @@
...
+      return(data.frame(lower = na, upper = na))
+    }
+  }
+
+  x <- sort(x)
+  n <- length(x)
+
+  if (n <= 1) {
+    return(data.frame(lower = na, upper = na))
+  }
```
joethorley:
```suggestion
  n <- length(x)

  if (n <= 1) {
    return(data.frame(lower = na, upper = na))
  }
  x <- sort(x)
```

fractional speed up

R/rope.R
```diff
@@ -0,0 +1,56 @@
...
+                     type = "HDI", na_rm = FALSE) {
+  chk_numeric(x)
+  chk_number(threshold)
+  chk_numeric(interval)
+  chk_length(interval, 2)
+  chk_not_any_na(interval)
+  chk_sorted(interval)
+  chk_unused(...)
+  chk_number(level)
+  chk_range(level)
```
joethorley, line 43:
```suggestion
  chk_range(level)
  chk_gt(level)
```

presumably we need level to be greater than 0?

joethorley (changes requested):
some refactoring for your review

its coming together in some very tight functions

R/rope.R
StefanoMezzini: `level > 0` is tested downstream by the CI functions

R/ci_hdi.R
StefanoMezzini: skipping since tidyverse prefers expanded format

joethorley (dismissed):
looks good

R/ci.R
```diff
@@ -0,0 +1,26 @@
...
+#' coverage of the CI.
+#' @param ... Currently unused.
+#' @param na_rm A flag indicating whether to remove missing values.
+#' @return A data frame of the `lower` and `upper` limits for the credible interval.
+#' Note that the interval is not guaranteed to be one-sided or two-sided.
+#' @export
+#' @seealso [mcmcr::coef]
+#' @examples
+#' xtr_ci(rnorm(1e4), type = "HDI")
+xtr_ci <- function(x, level = 0.95, ..., type = "HDI", na_rm = FALSE) {
```
aylapear, line 20: Missing argument documentation for `type`

R/ci.R
```diff
@@ -0,0 +1,26 @@
...
+  chk_subset(type, c("HDI", "ETI"))
+  chk_unused(...)
+
+  if(type == "HDI") {
```
aylapear: Minor but there should be a space between the if and the bracket. 


```suggestion
  if (type == "HDI") {
```

R/ci.R
aylapear, line 1: If there is only 1 function in the file then the best practice is to have the file named the same as the function so this file name would be *xtr_ci*

R/ci_hdi.R
```diff
@@ -0,0 +1,89 @@
...
+#' xtr_ci_hdi(1:100) # inclusive interval [3, 98] with 5% of values outside
+xtr_ci_hdi <- function(x, level = 0.95, ..., na_rm = FALSE) {
+  chk_numeric(x)
+  chk_number(level)
+  chk_range(level, inclusive = TRUE)
+  chk_gt(level)
+  chk_unused(...)
+  chk_flag(na_rm)
+
+  if(is.integer(x)) {
```
aylapear:
```suggestion
  if (is.integer(x)) {
```

R/ci_hdi.R
```diff
@@ -0,0 +1,89 @@
...
+    if (n_inf <= n_out) {
+      actual_n_ins <- sapply(narrowest_is, function(.i) {
+        .l <- x[.i + n_in - 1]
+        .u <- x[.i]
+        sum(x >= .l & x <= .u)
+      })
+      narrowest_is <- narrowest_is[actual_n_ins == max(actual_n_ins)]
+      midpoints <- narrowest_is + (n_in - 1) / 2
+      narrowest_i <- narrowest_is[which.min(abs(midpoints - (1 + n) / 2))]
+    } else if(n_inf < n_in) {
```
aylapear:
```suggestion
    } else if (n_inf < n_in) {
```

R/ci_hdi.R
aylapear, line 1: Update file name to reflect function name

R/ci_eti.R
aylapear, line 1: Update file name to reflect function name

R/ci_eti.R
```diff
@@ -0,0 +1,47 @@
+#' Equal-Tailed Interval
+#'
+#' Calculates Bayesian credible intervals using the equal-tailed interval (ETI),
+#' i.e., the CI such that the left and right tails outside the CI have the same
+#' coverage. Note that the function does not return integer outputs, even if the
+#' input data are integers, unlike [`xtr_ci_hdi()`].
```
aylapear: This note about the return I would put in the return section or the details of the function.

R/ci_eti.R
```diff
@@ -0,0 +1,47 @@
...
+#'
+#' @param x A numeric vector of MCMC samples.
+#' @param level A number > 0 and <= 1 specifying the probability coverage of the
+#' ETI.
+#' @param ... Currently unused.
+#' @param na_rm A flag indicating whether to remove missing values.
+#' @return A [data.frame] of the `lower` and `upper` limits for the credible
+#' interval.
+#' Note that the interval is guaranteed to be two-sided with real (i.e., double)
+#' numeric limits, even if the input data are integers.
```
aylapear: This is kind of duplicated information in regards to the function outputting a double. Something between this and the other note about should be combined so it is a single statement that describes the behaviour of the return. 

I would suggest something like this

```
#' @return A [data.frame] with columns `lower` and `upper` limits for the credible interval.

#' @details
#' Does not return integer outputs even if the input data are integers, unlike [`xtr_ci_hdi()`]. The interval limits are always real (double) numeric values.
```

R/ci_eti.R
```diff
@@ -0,0 +1,47 @@
...
+#' xtr_ci_eti(rnorm(1e4))
+xtr_ci_eti <- function(x, level = 0.95, ..., na_rm = FALSE) {
+  chk_numeric(x)
+  chk_number(level)
+  chk_range(level, inclusive = TRUE)
+  chk_gt(level)
+  chk_unused(...)
+  chk_flag(na_rm)
+
+na <-  if(is.integer(x)) NA_integer_ else NA_real_
```
aylapear:
```suggestion
na <-  if (is.integer(x)) NA_integer_ else NA_real_
```

R/ci_eti.R
```diff
@@ -0,0 +1,47 @@
...
+
+  if(length(x) <= 1) {
```
aylapear:
```suggestion
  if (length(x) <= 1) {
```

R/pvalue.R
```diff
@@ -1,13 +1,23 @@
...
-#' By default a p-value of 0.05 indicates that the 95% CRI just includes 0.
+#' By default a p-value of 0.05 indicates that the 95% CRI just includes the
+#' threshold value.
+#'
+#' Note that the function contains the sample-size correction
+#' \eqn{p_{c} = p * n / (n + 1)} to avoid p-values of 0. The function can still
+#' return p-values of 1.
+#'
+#' For p-values converted to bits, see [`svalue()`].
+#' }
```
aylapear: I think this should all go in the details section of the docs.

R/rope.R
aylapear, line 1: Update file name to reflect function name

R/rope.R
```diff
@@ -0,0 +1,56 @@
...
+#' @param na_rm A flag indicating whether to remove missing values.
+#' @return A number indicating the estimated proportion of the posterior (or
+#' credible interval) within the region of practical equivalence.
+#' @export
+#' @seealso [extras::xtr_ci]
+#' @examples
+#' xtr_rope(c(-Inf, -1, -0.1, 0.1, 1, Inf))
+#' xtr_rope(rnorm(1e4))
+#' @name xtr_rope
+NULL
```
aylapear: This can be removed.

R/rope.R
```diff
@@ -0,0 +1,56 @@
...
+#' @param ... Currently unused.
+#' @param na_rm A flag indicating whether to remove missing values.
+#' @return A number indicating the estimated proportion of the posterior (or
+#' credible interval) within the region of practical equivalence.
+#' @export
+#' @seealso [extras::xtr_ci]
+#' @examples
+#' xtr_rope(c(-Inf, -1, -0.1, 0.1, 1, Inf))
+#' xtr_rope(rnorm(1e4))
+#' @name xtr_rope
```
aylapear: I think this can also be removed as we normally don't have this field.

R/rope.R
```diff
@@ -0,0 +1,56 @@
...
+NULL
+
+xtr_rope <- function(x, threshold = 0, interval = c(-0.1, 0.1), ..., level = 1,
```
aylapear, line 33: It looks like `threshold`, `interval` and `type` are missing an argument descriptions.

aylapear, line 33: It appears the usage section is missing from the rendered docs when I tried on my computer. 

Image

R/rope.R
```diff
@@ -0,0 +1,56 @@
...
+    } else {
+      NA_real_
+    }
+  }
+
+  if (length(x) < 2) {
+    return(NA_real_)
+  }
+
+  if(level < 1) {
```
aylapear:
```suggestion
  if (level < 1) {
```

tests/testthat/test-ci-eti.R
aylapear, line 1: test file name should be updated to reflect the function name

tests/testthat/test-ci-hdi.R
aylapear, line 1: Test file name should reflect the function name

tests/testthat/test-ci.R
aylapear, line 1: update test file name to reflect function name

tests/testthat/test-rope.R
aylapear, line 1: test file name should reflect function name

tests/testthat/test-ci-eti.R
```diff
@@ -0,0 +1,30 @@
...
+  expect_equal(xtr_ci_eti(c(NA_integer_, 1:2), na_rm = TRUE), data.frame(lower = 1.025, upper = 1.975))
+  expect_error(xtr_ci_eti(1, 1, TRUE), "`...` must be unused.")
+
+  expect_no_error(xtr_ci_eti(matrix(1:9, ncol = 3)))
+  expect_no_error(xtr_ci_eti(array(1:27, dim = c(3, 3, 3))))
+  expect_error(xtr_ci_eti(list(1)), "`x` must be numeric.")
+  expect_error(xtr_ci_eti(data.frame(1)), "`x` must be numeric.")
+
+  expect_error(xtr_ci_eti(1:100, level = 0), "level` must be greater than 0, not 0.")
+  expect_no_error(xtr_ci_eti(1:100, level = 1))
```
aylapear, line 17: I would add one test to confirm it errors when greater than 1

```
xtr_ci_eti(1:100, level = 1.1)
```

tests/testthat/test-ci-eti.R
```diff
@@ -0,0 +1,30 @@
...
+
+  expect_equal(xtr_ci_eti(1:100, level = 1), data.frame(lower = 1, upper = 100))
+  expect_equal(xtr_ci_eti(NA_real_), data.frame(lower = NA_real_, upper = NA_real_))
```
aylapear: This is a duplicate as exact same code as line 2

tests/testthat/test-ci-eti.R
```diff
@@ -0,0 +1,30 @@
...
+  expect_equal(xtr_ci_eti(numeric(0)), data.frame(lower = NA_real_, upper = NA_real_))
```
aylapear: This is a duplicate test as exact same code as line 4

tests/testthat/test-ci-hdi.R
```diff
@@ -0,0 +1,94 @@
...
+
+  expect_no_error(xtr_ci_hdi(matrix(1:9, ncol = 3)))
+  expect_no_error(xtr_ci_hdi(array(1:27, dim = c(3, 3, 3))))
+  expect_error(xtr_ci_hdi(list(1)), "`x` must be numeric.")
+  expect_error(xtr_ci_hdi(data.frame(1)), "`x` must be numeric.")
+
+  expect_error(xtr_ci_hdi(1:100, level = 0), "level` must be greater than 0, not 0.")
+  expect_no_error(xtr_ci_hdi(1:100, level = 1))
+
+  # differences between small even and odd sample sizes
```
aylapear, line 21: Since you are adding comments to explain what the different test blocks are doing I would switch them to their own test_that block. The tests in this package are generally all in 1 test_that block but there are a lot in this file and the comments show that there are clearly some themes that I think are worth splitting into their own block.

tests/testthat/test-ci-hdi.R
```diff
@@ -0,0 +1,94 @@
...
+  expect_error(xtr_ci_hdi(1, 1, TRUE), "`...` must be unused.")
+
+  expect_no_error(xtr_ci_hdi(matrix(1:9, ncol = 3)))
+  expect_no_error(xtr_ci_hdi(array(1:27, dim = c(3, 3, 3))))
+  expect_error(xtr_ci_hdi(list(1)), "`x` must be numeric.")
+  expect_error(xtr_ci_hdi(data.frame(1)), "`x` must be numeric.")
+
+  expect_error(xtr_ci_hdi(1:100, level = 0), "level` must be greater than 0, not 0.")
+  expect_no_error(xtr_ci_hdi(1:100, level = 1))
+
```
aylapear, line 19: Add a test for when level is greater than 1.

tests/testthat/test-ci.R
```diff
@@ -0,0 +1,5 @@
+test_that("xtr_ci() works with all CI types.", {
+  expect_no_error(xtr_ci(1:100))
+  expect_no_error(xtr_ci(1:100, type = "HDI"))
+  expect_no_error(xtr_ci(1:100, type = "ETI"))
```
aylapear, line 4: Add a test to confirm that type errors if not one of the two values it expects.

R/ci.R
```diff
@@ -0,0 +1,26 @@
...
+#' @param ... Currently unused.
+#' @param na_rm A flag indicating whether to remove missing values.
+#' @return A data frame of the `lower` and `upper` limits for the credible interval.
+#' Note that the interval is not guaranteed to be one-sided or two-sided.
+#' @export
+#' @seealso [mcmcr::coef]
+#' @examples
+#' xtr_ci(rnorm(1e4), type = "HDI")
+xtr_ci <- function(x, level = 0.95, ..., type = "HDI", na_rm = FALSE) {
+  chk_subset(type, c("HDI", "ETI"))
```
aylapear, line 23: Does this also need a `chk_string()`? I feel like we had a convo about this in another PR somewhere but I can't recall. 

This was done in the `pvalue()` function, I feel like there was a reason why but maybe I am just mixing it up with something else.

aylapear (changes requested):

aylapear:
I am just reviewing the code but haven't thought through the logic and actual correctness of the function. @nehill197 I am assuming you and Joe are thinking more about the actual corrections of the function in your review.

joethorley:
@aylapear - yes I focused on the correctness of the functions but it would be good to have Nicole double check.

R/rope.R
aylapear: Sorry only the NULL and @name part should have been highlighted.

R/rope.R
aylapear: Only the @name tag should have been highlighted.

R/ci.R
StefanoMezzini: fixed in 741c98f98cd907885e86701ec1adc12c2cfbfef1

R/ci.R
StefanoMezzini: fixed in ee59b194f86d613242a56c279d6192088849c200

R/ci.R
StefanoMezzini: I agree, but none of the other functions follow that naming convention (e.g., `xtr_mean()`, `xtr_median()`), so I was trying to stick with the package convention.

R/ci.R
StefanoMezzini: i think it does, because `vld_subset()` does not check for variable type:
```
> chk::vld_subset(1, letters)
[1] FALSE
```

R/ci_hdi.R
StefanoMezzini: not done because the other functions in the package don't follow that convention

R/ci_eti.R
StefanoMezzini: not done because the other functions in the package don't follow that convention

R/ci.R
StefanoMezzini: added the check in commit 237e19e66c9f48c783e1f7698514de6362ff7584

StefanoMezzini: and the test in 20826621c8802133835444b768bf6bfa348da236

R/ci_eti.R
StefanoMezzini: fixed in 20826621c8802133835444b768bf6bfa348da236

R/rope.R
StefanoMezzini: not done since the rest of the package does not follow this convention

R/pvalue.R
StefanoMezzini: fixed in fd06c8a41659c90c8cb22f59ef6dc55a8b1e847b

R/rope.R
StefanoMezzini: fixed in ea053b582b18b3fa8ad0b7adc9853859a43cbaf4

R/rope.R
StefanoMezzini: fixed in ea053b582b18b3fa8ad0b7adc9853859a43cbaf4

R/rope.R
StefanoMezzini: fixed in e2c4d56d28bc3dbacad002d14fa629c7d536fe1c

StefanoMezzini: fixed in 681d6365e2c7adfbeab6757f3aef3252a0627e02

tests/testthat/test-ci-eti.R
StefanoMezzini: not done since the rest of the package does not follow this convention

R/ci_eti.R
StefanoMezzini: fixed in 4cbcbf2ea1ba69025ff11ebbe5a1885deb997e40

tests/testthat/test-ci-hdi.R
StefanoMezzini: not done since the rest of the package does not follow this convention

tests/testthat/test-ci.R
StefanoMezzini: not done since the rest of the package does not follow this convention

tests/testthat/test-rope.R
StefanoMezzini: not done since the rest of the package does not follow this convention

tests/testthat/test-ci-eti.R
StefanoMezzini: fixed in 20826621c8802133835444b768bf6bfa348da236

tests/testthat/test-ci-eti.R
StefanoMezzini: fixed in 436d22c4ca5faff205e8f8424037e1d7c83638bd

R/ci.R
```diff
@@ -0,0 +1,30 @@
+#' Credible Intervals
+#'
+#' Calculates Bayesian credible intervals (CI) using one of the available methods:
+#' - `"HDI"`: highest density interval (see [xtr_ci_hdi()]),
+#' - `"ETI"`: equal tailed intervals (see [xtr_ci_eti()]).
+#'
+#' @param x A numeric vector of MCMC samples.
+#' @param level A number between 0 and 1 (exclusive) specifying the probability
```
nehill197: Why is this different to the parameter descriptions below for the functions for the individual intervals? (i.e., this excludes 1, while the ones below include 1). Consider adding this param to params.R?

R/ci_eti.R
```diff
@@ -0,0 +1,51 @@
...
+#' xtr_ci_eti(rnorm(1e4))
+xtr_ci_eti <- function(x, level = 0.95, ..., na_rm = FALSE) {
+  chk_numeric(x)
+  chk_number(level)
+  chk_range(level, inclusive = TRUE)
+  chk_gt(level)
+  chk_unused(...)
+  chk_flag(na_rm)
+
+na <-  if (is.integer(x)) NA_integer_ else NA_real_
```
nehill197: some weird formatting here - can you indent to line up with the rest of the function and remove the extra space after the arrow?

R/ci_hdi.R
```diff
@@ -0,0 +1,89 @@
...
+#' interval.
+#' Note that the interval is not guaranteed to be one-sided or two-sided.
+#' Returns integer limits if the input data are integers and double otherwise.
+#' @export
+#' @seealso [extras::xtr_ci()] and [extras::xtr_ci_eti()]
+#' @examples
+#' xtr_ci_hdi(1:10, level = 0.1) # only 10% of values inside
+#' xtr_ci_hdi(1:10, level = 0.2) # only 20% of values inside
+#' xtr_ci_hdi(1:10, level = 0.2 + 0.01) # at least 20.1% of values inside
+#' xtr_ci_hdi(1:100) # inclusive interval [3, 98] with 5% of values outside
```
nehill197: I'd find it (slightly) easier to understand if you say 95% of values inside to parallel the other examples.

R/ci.R
```diff
@@ -0,0 +1,30 @@
...
+#' Currently allows Highest Density Intervals (`"HDI"`; default) and
+#' Equal-Tailed Intervals (`"ETI"`).
+#' @param na_rm A flag indicating whether to remove missing values.
+#' @return A data frame of the `lower` and `upper` limits for the credible interval.
+#' Note that the interval is not guaranteed to be one-sided or two-sided.
+#' @export
+#' @seealso [mcmcr::coef]
+#' @examples
+#' xtr_ci(rnorm(1e4), type = "HDI")
+xtr_ci <- function(x, level = 0.95, ..., type = "HDI", na_rm = FALSE) {
```
nehill197, line 20: missing chk_flag() for `na_rm`?

R/ci_hdi.R
```diff
@@ -0,0 +1,89 @@
...
+  chk_flag(na_rm)
+
+  if (is.integer(x)) {
+    na <- NA_integer_
+  } else {
+    na <- NA_real_
+  }
+
+  if (anyNA(x)) {
+    if (vld_true(na_rm)) {
```
nehill197: why using `vld_true(na_rm)` here and not for the other functions that have this same construction?

R/rope.R
```diff
@@ -0,0 +1,61 @@
...
+#' is sensitive to unit conversions.
+#'
+#' @param x A numeric vector of MCMC samples.
+#' @param threshold A number specifying the center of the ROPE.
+#' @param interval A numeric vector of length 2 to be added to `threshold` to
+#' calculate the ROPE.
+#' Generally, `threshold[1] == - threshold[2]` and `threshold[1] < 0` should
+#' both be true.
+#' @param level A number > 0 and <= 1 specifying the probability coverage of
+#' the HDI to use.
```
nehill197: if you can use ETI intervals instead I think we should just say the interval here, not HDI.

R/rope.R
```diff
@@ -0,0 +1,61 @@
...
+#' The default of 1 uses the full posterior.
+#' @param ... Currently unused.
+#' @param type A string indicating which type of CI to return.
+#' Currently allows Highest Density Intervals (`"HDI"`; default) and
+#' Equal-Tailed Intervals (`"ETI"`).
+#' @param na_rm A flag indicating whether to remove missing values.
+#' @return A number indicating the estimated proportion of the posterior (or
```
nehill197: Is it actually estimated??

R/rope.R
```diff
@@ -0,0 +1,61 @@
...
+  chk_unused(...)
+  chk_number(level)
+  chk_range(level)
+  chk_flag(na_rm)
+
+  if (anyNA(x)) {
+    if (vld_true(na_rm)) {
+      x <- x[!is.na(x)]
+    } else {
+      NA_real_
```
nehill197: Don't think this line is doing anything - presumably you meant to have `return(NA_real_)`?

R/ci_hdi.R
```diff
@@ -0,0 +1,89 @@
...
+  }
+  x <- sort(x)
+
+  if (level == 1) {
+    return(data.frame(lower = x[1], upper = x[n]))
+  }
+
+  n_in <- ceiling(n * level)
+  n_out <- n - n_in
+  n_inf <- sum(is.infinite(x))
```
nehill197, line 56: add infinity checks here
(vaguely following this)

if (n_inf == n) {
   if (sum(is.infinite(x) & x > 0) {
     return(c(Inf, Inf))
   } else if (sum(is.infinite(x)) & x < 0) {
     return(c(-Inf, -Inf))
   } else { 
     return(c(-Inf, Inf))
  }
}

R/ci_hdi.R
```diff
@@ -0,0 +1,89 @@
...
+      midpoints <- narrowest_is + (n_in - 1) / 2
+      narrowest_i <- narrowest_is[which.min(abs(midpoints - (1 + n) / 2))]
+    } else if (n_inf < n_in) {
+      if (is.infinite(x[1])) {
+        narrowest_i <- min(narrowest_is)
+      } else {
+        narrowest_i <- max(narrowest_is)
+      }
+    } else {
+      return(data.frame(lower = x[1], upper = x[n]))
```
nehill197, line 91: check this would meet the criteria if we have x = c(-Inf, rep(Inf, 19))

nehill197 (changes requested):
some comments in here to look at!

R/ci.R
StefanoMezzini: fixed in ac9586e

R/ci_eti.R
StefanoMezzini: fixed in ed4e7f5f2f5138c42650466f8ddff7fc8b248811

R/ci_hdi.R
StefanoMezzini: fixed in f6fbd787c1bb1364fcaa914b7a48744f52e3f0a7

R/ci.R
StefanoMezzini: fixed in 80ba9ac9ab43de739e549cd7f05854427ec7ce9d

R/ci_hdi.R
StefanoMezzini: I was initially using `vld_true()` because it was in other functions, but Joe pointed out that it's redundant if I've already used `chk_flag()`, so I've switched to using `chk_flag()` throughout. Fixed in 80ba9ac9ab43de739e549cd7f05854427ec7ce9d and a2cb274e4809551ea68ed9e1980a803af41a9f1a

R/rope.R
StefanoMezzini: fixed in 46e7f118d7dd0494808927ca08c4f95b0fb07715

R/rope.R
StefanoMezzini: no, it's calculated -- fixed in 393b4174542f6a7636006424f52377898f5138a8

I got the description mixed up because the posterior is estimating the truth

R/rope.R
StefanoMezzini: fixed in afcec75e65e334c3d8e79e17d6155b5e68be51c9

R/ci_hdi.R
StefanoMezzini: fixed in 9ce089882c0bae5f94f34fb1726e7a2d955000a4, but this didn't deal with situations like `c(rep(-Inf, 19), Inf)` correctly, so I made some changes in d2624b3bc1f8ed84affe7f9936ca0b6d109a53f0

R/ci_hdi.R
StefanoMezzini: fixed in d2624b3bc1f8ed84affe7f9936ca0b6d109a53f0

StefanoMezzini:
thanks for catching those issues and going through it with me! I'm happier with the functions now, and I feel more confident in how it deals with `+/-Inf` values. I've checked the tests for `xtr_ci_hdi()` again, too.

R/params.R
```diff
@@ -17,6 +17,8 @@
 #' @param directional A flag specifying whether probabilities less than 0.5
 #'   should be returned as negative values.
 #' @param lambda A non-negative numeric vector of means.
+#' @param level A number between 0 and 1 (inclusive) specifying the probability
```
nehill197: I think I left a comment on the commit so sorry if you're getting this twice but I'm pretty sure the functions don't let `level` be 0! So > 0 and <= 1

nehill197 (changes requested):
One more minor thing!

R/params.R
StefanoMezzini: good catch! fixed in f163d877146c9495b6af9d7246f6e08bcfbf0fa3.

StefanoMezzini: i was initially allowing `level = 0`, but I don't think it makes sense. setting `level = 1` is essentially an overly complicated `range()`, but it still makes some sense.

R/params.R
```diff
@@ -17,6 +17,8 @@
 #' @param directional A flag specifying whether probabilities less than 0.5
 #'   should be returned as negative values.
 #' @param lambda A non-negative numeric vector of means.
+#' @param level A number > 0 and <= 1 (inclusive) specifying the probability
```
nehill197: Okay so close but I think that inclusive is redundant here if we're explicitly stating that 0 is excluded and 1 is included.

nehill197 (changes requested):

R/params.R
StefanoMezzini: forgot to remove it 🤦 fixed in d7a07d2fc9252cc77f71c80c66f6760c81b8b147

nehill197 (dismissed):
Nice!!

R/pvalue.R
aylapear: Since the `{` was deleted on line 3 it has created a mismatch. Please correct and re-build the docs.

R/ci.R
```diff
@@ -0,0 +1,30 @@
...
+#' - `"ETI"`: equal tailed intervals (see [xtr_ci_eti()]).
+#'
+#' @inheritParams params
+#' @param x A numeric vector of MCMC samples.
+#' @param ... Currently unused.
+#' @param type A string indicating which type of CI to return.
+#' Currently allows Highest Density Intervals (`"HDI"`; default) and
+#' Equal-Tailed Intervals (`"ETI"`).
+#' @param na_rm A flag indicating whether to remove missing values.
+#' @return A data frame of the `lower` and `upper` limits for the credible interval.
```
aylapear:
```suggestion
#' @return A [data.frame] of the `lower` and `upper` limits for the credible interval.
```

Updating to be consistent with other `ci` functions return statement.

aylapear (changes requested):

R/pvalue.R
StefanoMezzini: fixed in 3550a28

aylapear (approved):
Thanks for fixing those last things up.

joethorley (approved):

nehill197 (approved):

PR #111, opened 2026-05-06
fix issue with `renv`

StefanoMezzini:


aylapear:
This PR should be closed as renv lock files are not used for our packages. The issue is related to `tibble` being called in the code but is incorrectly listed in the DESCRIPTION file. Since `tibble` is used inside a function it needs to be listed in Imports but it is listed in Suggests. 

If you update that it should fix the error. We don't get the error locally because we already have tibble installed but the GitHub action machine doesn't install tibble since it is listed as a suggests and thus we get an error when the example is called. This is also why we get the error when the tests for those functions are called. I would close this PR and then make the DESCRIPTION file change in this PR https://github.com/poissonconsulting/extras/pull/110.

aylapear (changes requested):
See comment above

StefanoMezzini:
fixed the R CMD check in #110 via commit f58889faac8ed5463e9b7532bf023c653d59aaaf

PR #122, opened 2026-06-12
- Add functions to calculate s-values and information from vectors of probabilities

### LINKED ISSUE #121: Add functions to calculate s-values and directional information from probabilities
`svalue()` and `directional_information()` require a vector of MCMC samples, so we currently don't have functions that estimate the two metrics directly from probabilities. I think it would be useful to have some exported functions that perform those calculations. We could name them something like `pvalue_2_svalue()` and `pvalue_2_dinfo()` or even `p_2_s()` and `p_2_di()`.

I am happy to make the PR for this.

StefanoMezzini:
fixes https://github.com/poissonconsulting/extras/issues/121


nehill197 (dismissed):
Looks good!

StefanoMezzini:
again, the dismissal of the stale review seems to be an automatic consequence of merging in the main branch

joethorley (commented):
This is great.

I'm only knit-picky this is the naming of functions

`p_2_info()`
`p_2_svalue()`

I definitely prefer

`p2info()`
`p2svalue()`

@nehill197 - what do you as CODEOWNER think?

nehill197:
I mildly agree with you so yeah let's switch it!

joethorley (approved):

nehill197 (approved):
Looks good!

PR #133, opened 2026-07-10
Add skeptical argument to directional_information(), pvalue(), and svalue()

nehill197:
## Summary

- Adds `skeptical` argument to `directional_information()`: when `TRUE` and all samples are on one side of the threshold, adds one sample to the opposite side instead of clamping to `±n`, giving `sign(i) * log2(n)`
- Adds `skeptical` argument to `pvalue()` and `svalue()` 
- Adds `...` with `chk_unused()` to all three functions to catch unrecognised arguments
- Adds examples and tests for all new behaviour

## Test plan

- [x] `testthat::test_file("tests/testthat/test-directional-information.R")` passes
- [x] `testthat::test_file("tests/testthat/test-pvalue.R")` passes with no warnings
- [x] `testthat::test_file("tests/testthat/test-svalue.R")` passes with no warnings


nehill197:
just had a thought that maybe the `skeptical = FALSE` for directional_information() should be Infinity, not the sample size? I like the sample size as an adjustment but it really is a middle ground.. let me know what you think when you take a look @joethorley

joethorley (approved):
looks great

StefanoMezzini:
I was thinking that keeping some type of lower limit for p-values would be helpful, but even if we set it to be $2^{-n}$ to match the limit on information of $\pm n$ we would effectively still only be applying the limit to samples with $n <= 52$, since `.Machine$double.eps`$^*$ would be the limit for a sample size of $n = -\log_2(2.220446\times10^{-16}) = 52$, which is well below what we would use for MCMC samples, anyway.

So I agree that it makes sense to drop the limits for p-values and s-values.

------

$^*$ `.Machine$double.eps` is `2.220446e-16`, supposedly the smallest positive floating-point number `e` such that `1 + e != 1` in `R`, although I found that the number is actually closer to `1+1.11e-16`$\approx 2^{-53}$:
``` r
> 1+.Machine$double.eps==1
[1] FALSE
> 1+1.111e-16==1
[1] FALSE
> 1+1.11e-16==1
[1] TRUE
> p2svalue(1.11e-16)
[1] 53.00029
```

StefanoMezzini:
I do like keeping the limits for the information though because it keeps it within a framework of maximum entropy for outcomes: With no prior knowledge (or assuming most entropic case), the probability that a sample be on one side of the threshold is 0.5 (since the variance of a Bernoulli random variable, $\sigma^2=pq$, is maximized when $p=0.5$). This implies that the most information we can gain from a sample of size $n$ is equal to $n$ bits. Conceptually, I also find it satisfying to think of a stack of $n$ coins (or plinko with a single split) sitting on one side of the threshold and representing the full sample.

nehill197:
Sounds reasonable to me @StefanoMezzini!

R/directional-information.R
```diff
@@ -29,6 +30,13 @@
 #' @param n A numeric vector of the number of posterior samples used to estimate
 #' each value of `p`. Used to limit the information to be within the interval \eqn{[-n, n]}.
 #'
+#' @details
+#' When `skeptical = FALSE` (default), information is clamped to \eqn{[-n, n]}
+#' when all samples are on one side of the threshold.
+#' When `skeptical = TRUE`, one sample is instead added to the empty side,
+#' giving \eqn{\pm \log_2(n)} rather than \eqn{\pm n}, to acknowledge additional
+#' uncertainty.
```
StefanoMezzini: what's the rationale for bounding it to $\pm \log_2(n)$, other than doing what we do for p-values?

tests/testthat/test-directional-information.R
```diff
@@ -44,6 +44,34 @@ test_that("directional_information() returns correct values.", {
                4.246, tolerance = 0.01)
 })
 
+test_that("directional_information() errors with unused arguments", {
+  expect_error(directional_information(1, 0))
+  expect_error(directional_information(1, foo = TRUE))
```
StefanoMezzini: to ensure that the error message is the one expected

```suggestion
  expect_error(directional_information(1, 0),  "`...` must be unused.")
  expect_error(directional_information(1, foo = TRUE),  "`...` must be unused.")
```

tests/testthat/test-directional-information.R
```diff
@@ -44,6 +44,34 @@ test_that("directional_information() returns correct values.", {
...
+})
+
+test_that("directional_information() skeptical argument works correctly", {
```
StefanoMezzini:
```suggestion
test_that("directional_information() skeptical argument works correctly with default side", {
```

tests/testthat/test-pvalue.R
```diff
@@ -1,117 +1,141 @@
...
-                            threshold = qpois(0.05, 3)),
+                            threshold = qpois(0.05, 3), skeptical = TRUE),
                ppois(qpois(0.05, 3) - 1, # correct for boundary inclusion
                            3, lower.tail = FALSE), tolerance = 0.01)
 
 })
 
+test_that("pvalue() errors with unused arguments", {
+  expect_error(pvalue(1, 0))
+  expect_error(pvalue(1, foo = TRUE))
```
StefanoMezzini: to ensure that the error message is the one expected
```suggestion
  expect_error(pvalue(1, 0), "`...` must be unused.")
  expect_error(pvalue(1, foo = TRUE), "`...` must be unused.")
```

tests/testthat/test-svalue.R
```diff
@@ -1,35 +1,59 @@
...
+  expect_equal(svalue(-9:10 + 1, side = "right", threshold = 1, skeptical = TRUE), -log2(11 / 20))
+  expect_identical(svalue(c(1, 1, NA), side = "left", skeptical = TRUE), NA_real_)
+  expect_equal(svalue(c(1, 1, NA), side = "left", na_rm = TRUE, skeptical = TRUE), -log2(1 / 3))
+  expect_identical(svalue(-c(1, 1, NA), side = "right", skeptical = TRUE), NA_real_)
+  expect_equal(svalue(-c(1, 1, NA), side = "right", na_rm = TRUE, skeptical = TRUE), -log2(1 / 3))
+})
+
+test_that("svalue() errors with unused arguments", {
+  expect_error(svalue(1, 0))
+  expect_error(svalue(1, foo = TRUE))
```
StefanoMezzini:
```suggestion
  expect_error(svalue(1, 0), "`...` must be unused.")
  expect_error(svalue(1, foo = TRUE), "`...` must be unused.")
```

tests/testthat/test-svalue.R
```diff
@@ -1,35 +1,59 @@
...
+  expect_identical(svalue(c(-1, 1, 1), skeptical = TRUE),
+                   svalue(c(-1, 1, 1), skeptical = FALSE))
+  expect_identical(svalue(c(-1, 1, 1), side = "left", skeptical = TRUE),
+                   svalue(c(-1, 1, 1), side = "left", skeptical = FALSE))
+  expect_identical(svalue(c(-1, 1, 1), side = "right", skeptical = TRUE),
+                   svalue(c(-1, 1, 1), side = "right", skeptical = FALSE))
 })
 
 test_that("svalue() requires side to be one of left, right, both, or NULL, and returns a warning with default side.", {
-  expect_no_error(pvalue(1, side = "both"))
```
StefanoMezzini, line 29: Of course I missed changing `pvalue()` to `svalue()` when writing the tests 🤦

StefanoMezzini (changes requested):
looks good! just some minor edits

R/directional-information.R
nehill197: I think they both are doing the same thing - adding a phantom sample to one side:

for pvalue:
```
p = 1 / (n + 1)
```

for directional info - if samples are all on right side:
```
p_r = n/(n+1),  p_l = 1/(n+1)
i = log2(p_r) - log2(p_l)
  = log2(n/(n+1)) - log2(1/(n+1))
  = log2(n/(n+1) / (1/(n+1)))
  = log2(n)
```

StefanoMezzini: yes, what I mean is there any reason someone would set it to `TRUE`, other than for consistency with `pvalue()`? I'm just wondering whether we should add any info to the description.

StefanoMezzini: what I'm hung up on is the "to acknowledge additional uncertainty", since reading that immediately makes me wonder where that uncertainty is coming from, since that text doesn't explain that. Maybe it's just the verb "acknowledge" that is tripping me up, although "to add additional uncertainty" wouldn't be any better

StefanoMezzini: resolved this in a huddle

R/directional-information.R
```diff
@@ -110,12 +123,16 @@ directional_information <- function(x, side = "median", threshold = 0,
...
+  if (skeptical && is.infinite(i)) {
+    i <- sign(i) * log2(n)
+  } else {
+    i <- min(i, n) # max information difference is a bit for each sample
+    i <- max(i, -n)
+    # the two lines above are equivalent to, if o is the odds ratio:
+    # o being n   if i is  Inf, since n    = n / (n+1) / (1 / (n+1))
+    # o being 1/n if i is -Inf, since 1/n  = 1 / (n+1) / (n / (n+1))
+    # note that o could be p_l / p_r or p_r / p_l
+  }
```
StefanoMezzini: If we are going to test for `is.infinite(i)`, we may as well wrap the `else` section into it, too, since it only applies when `is.infinite(i)` is `TRUE`

```suggestion
  if (is.infinite(i)) {
    if (skeptical) {
      i <- sign(i) * log2(n)
    } else {
      i <- min(i, n) # max information difference is a bit for each sample
      i <- max(i, -n)
      # the two lines above are equivalent to, if o is the odds ratio:
      # o being n   if i is  Inf, since n    = n / (n+1) / (1 / (n+1))
      # o being 1/n if i is -Inf, since 1/n  = 1 / (n+1) / (n / (n+1))
      # note that o could be p_l / p_r or p_r / p_l
    }
  }
```

man/directional-information.Rd
```diff
@@ -39,6 +43,10 @@ proportionally to the values of \code{x} on the left and right sides,
 (identical to using \code{"proportional"}).
 }}
 
+\item{skeptical}{A flag specifying whether or not to add one sample to the
+empty side of the threshold when 100\% of samples are on one side, to avoid
+infinite information or zero p-values.}
```
StefanoMezzini: the directional information is never infinite, since we bound it even when `skeptical = FALSE`. I've made edits in `params.R` and `directional-information.R`, but I avoided editing the docs directly to minimize typos

R/params.R
```diff
@@ -46,6 +46,9 @@
 #' @param shape A non-negative numeric vector of shape.
 #' @param simulate A flag specifying whether to simulate residuals.
 #' @param size A non-negative whole numeric vector of the number of trials.
+#' @param skeptical A flag specifying whether or not to add one sample to the 
+#' empty side of the threshold when 100% of samples are on one side, to avoid 
+#' infinite information or zero p-values.
```
StefanoMezzini: since the calculated directional info is never infinite:

```suggestion
#' zero p-values and infinite s-values, and also to impose stronger bounds on
#' directional information than $[-n, n]$, which assume the MCMC samples are
#' independent and representative.
```

tests/testthat/test-directional-information.R
```diff
@@ -2,8 +2,8 @@ test_that("directional_information() returns correct values.", {
   expect_identical(directional_information(NA_integer_), NA_real_)
   expect_identical(directional_information(NA_real_), NA_real_)
   expect_identical(directional_information(NaN), NA_real_)
-  expect_identical(directional_information(Inf), 1)
-  expect_identical(directional_information(-Inf), 1)
+  expect_identical(directional_information(Inf), 0)
+  expect_identical(directional_information(-Inf), 0)
```
StefanoMezzini:
```suggestion
  expect_identical(directional_information(Inf), log2(1))
  expect_identical(directional_information(-Inf), log2(1))
```

tests/testthat/test-directional-information.R
```diff
@@ -13,11 +13,11 @@ test_that("directional_information() returns correct values.", {
   expect_identical(directional_information(c(-1:2), threshold_split = "proportional"), log2(2/1))
   expect_identical(directional_information(c(-1:2), threshold_split = "exclude"), log2(2/1))
 
-  expect_identical(directional_information(1), 1)
-  expect_equal(directional_information(c(1, 1)), 2)
-  expect_equal(directional_information(c(1, 1, 1)), 3)
-  expect_equal(directional_information(1:9), 9)
-  expect_equal(directional_information(-(1:9)), 9)
+  expect_identical(directional_information(1), 0)
```
StefanoMezzini:
```suggestion
  expect_identical(directional_information(1), log2(1))
```

tests/testthat/test-pvalue.R
```diff
@@ -1,117 +1,135 @@
...
-  expect_equal(pvalue(c(0, 0)), 1)
-  expect_equal(pvalue(c(0, 0, 0)), 1)
-  expect_equal(pvalue(c(rep(-1, 25), rep(1, 1000 - 25))), 25 / 1000 * 2)
-  expect_equal(pvalue(c(rep(-1, 24), rep(1, 1000 - 24))), 24 / 1000 * 2)
-  expect_equal(pvalue(c(rep(-1, 25), rep(0, 1000 - 25))), (1000 - 25) / 1000)
-  expect_equal(pvalue(-9:10), 19/20)
-  expect_equal(pvalue(-9:10 + 1, threshold = 1), 19/20)
-  expect_equal(pvalue(c(-1, 1, 1, 1, NA)), NA_real_)
-  expect_equal(pvalue(c(-1, 1, 1, 1, NA), na_rm = TRUE), 1 / 4 * 2)
+  expect_identical(pvalue(NA_integer_, skeptical = TRUE), NA_real_)
```
StefanoMezzini: Since `skeptical = TRUE` is the default, I think you should drop "`, skeptical = TRUE`". It makes more sense to do a quick find and replace in Positron though rather than me suggesting all the changes here

StefanoMezzini: alternatively, you can revert to the previous file and then add the tests below back in

```r
test_that("pvalue() errors with unused arguments", {
  expect_error(pvalue(1, 0), "`...` must be unused.")
  expect_error(pvalue(1, foo = TRUE), "`...` must be unused.")
})

test_that("pvalue() skeptical argument controls sample-size correction", {
  expect_identical(pvalue(1:9, skeptical = TRUE), 1 / 10)
  expect_identical(pvalue(1:9, skeptical = FALSE), 0)
  expect_identical(pvalue(-(1:9), skeptical = TRUE), 1 / 10)
  expect_identical(pvalue(-(1:9), skeptical = FALSE), 0)
  expect_identical(pvalue(c(-1, 1, 1), skeptical = TRUE),
                   pvalue(c(-1, 1, 1), skeptical = FALSE))
  expect_identical(pvalue(c(-1, 1, 1), side = "left", skeptical = TRUE),
                   pvalue(c(-1, 1, 1), side = "left", skeptical = FALSE))
  expect_identical(pvalue(c(-1, 1, 1), side = "right", skeptical = TRUE),
                   pvalue(c(-1, 1, 1), side = "right", skeptical = FALSE))
})
```

tests/testthat/test-svalue.R
```diff
@@ -1,35 +1,53 @@
...
-  expect_equal(svalue(c(rep(1, 25), rep(-1, 1000 - 25)), side = "right"), -log2(25 / 1000))
-  expect_equal(svalue(-9:10, side = "left"), -log2(10 / 20))
-  expect_equal(svalue(-9:10, side = "right"), -log2(11 / 20))
-  expect_equal(svalue(-9:10 + 1, side = "left", threshold = 1), -log2(10 / 20))
-  expect_equal(svalue(-9:10 + 1, side = "right", threshold = 1), -log2(11 / 20))
-  expect_identical(svalue(c(1, 1, NA), side = "left"), NA_real_)
-  expect_equal(svalue(c(1, 1, NA), side = "left", na_rm = TRUE), -log2(1 / 3))
-  expect_identical(svalue(-c(1, 1, NA), side = "right"), NA_real_)
-  expect_equal(svalue(-c(1, 1, NA), side = "right", na_rm = TRUE), -log2(1 / 3))
+  expect_identical(svalue(integer(0), side = "left", skeptical = TRUE), NA_real_)
```
StefanoMezzini: same applies here, since `skeptical = TRUE` by default

R/directional-information.R
```diff
@@ -29,6 +30,13 @@
 #' @param n A numeric vector of the number of posterior samples used to estimate
 #' each value of `p`. Used to limit the information to be within the interval \eqn{[-n, n]}.
 #'
+#' @details
+#' When `skeptical = FALSE`, information is clamped to \eqn{[-n, n]}
+#' when all samples are on one side of the threshold.
+#' When `skeptical = TRUE` (default), one sample is instead added to the empty side,
+#' giving \eqn{\pm \log_2(n)} rather than \eqn{\pm n}, to mimic the behaviour of
+#' `pvalue()` and `svalue()`.
```
StefanoMezzini: I think it makes more sense to start with the default, especially since the non-skeptical interval is wider. Additionally, information is clamped to $[-n, n]$ even if it's not infinite, since it's still within the interval

```suggestion
#' When `skeptical = TRUE` (default), one sample is added to the empty side,
#' giving \eqn{\pm \log_2(n)} rather than \eqn{\pm n}, to mimic the behaviour of
#' `pvalue()` and `svalue()`.
#' When `skeptical = FALSE`, information is instead clamped to \eqn{[-n, n]},
#' which is assumes the MCMC samples are independent and representative.
```

again, will need to build locally to update the docs.

man/directional-information.Rd
```diff
@@ -57,6 +65,12 @@ information depending on the event's probability.
 \details{
 Quantifies the information about direction in a posterior distribution based on the directional probability.
 This function calculates such information using the difference in the probability of direction (see \code{\link[=probability_direction]{probability_direction()}}), after converting each probability to bits (also see \code{\link[=svalue]{svalue()}}.
+
+When \code{skeptical = FALSE}, information is clamped to \eqn{[-n, n]}
```
StefanoMezzini: see comment in the `R` script

R/svalue.R
```diff
@@ -17,10 +17,16 @@
 #' A near-impossible event has a very large s-value because its
 #' occurrence would be extremely surprising, like observing many consecutive
 #' successes on a fair coin.
+#'
+#' When `skeptical = TRUE` (default), a floor of \eqn{1 / (n + 1)} is applied
+#' to the underlying p-value to avoid s-values of `Inf` when all samples are on
```
StefanoMezzini:
```suggestion
#' When `skeptical = TRUE` (default), a ceiling of \eqn{n + 1} is applied
#' to the s-value to avoid s-values of `Inf` when all samples are on
```
will need to build locally to reflect changes in `Rd` file

StefanoMezzini (changes requested):

tests/testthat/test-pvalue.R
nehill197: fixed!

R/svalue.R
```diff
@@ -17,10 +17,16 @@
 #' A near-impossible event has a very large s-value because its
 #' occurrence would be extremely surprising, like observing many consecutive
 #' successes on a fair coin.
+#'
+#' When `skeptical = TRUE` (default), a ceiling of \eqn{n + 1} is applied
```
StefanoMezzini: I got confused with information when I wrote the previous correction 🤦 
```suggestion
#' When `skeptical = TRUE` (default), a ceiling of \eqn{\log_2(n + 1)} is applied
```

R/params.R
```diff
@@ -46,6 +46,11 @@
 #' @param shape A non-negative numeric vector of shape.
 #' @param simulate A flag specifying whether to simulate residuals.
 #' @param size A non-negative whole numeric vector of the number of trials.
+#' @param skeptical A flag specifying whether or not to add one sample to the 
+#' empty side of the threshold when 100% of samples are on one side, to avoid 
+#' zero p-values and infinite s-values, and also to impose stronger bounds on
```
StefanoMezzini:
```suggestion
#' empty side of the threshold when 100% of samples are on one side. Avoids 
#' zero p-values and infinite s-values, and also imposes stronger bounds on
```

R/directional-information.R
```diff
@@ -29,6 +30,13 @@
 #' @param n A numeric vector of the number of posterior samples used to estimate
 #' each value of `p`. Used to limit the information to be within the interval \eqn{[-n, n]}.
 #'
+#' @details
+#' When `skeptical = TRUE` (default), one sample is added to the empty side,
+#' giving \eqn{\pm \log_2(n)} rather than \eqn{\pm n}, to mimic the behaviour of
+#' `pvalue()` and `svalue()`.
```
StefanoMezzini:
```suggestion
#' giving bounds of \eqn{\pm \log_2(n)} rather than \eqn{\pm n}, to mimic the
#' behaviour of `pvalue()` and `svalue()`.
```

StefanoMezzini (changes requested):
Looks good! Just a last few minor fixes.

R/svalue.R
nehill197: haha good catch!

StefanoMezzini (approved):

PR #134, opened 2026-07-13
Fix info comments

StefanoMezzini:
fixed some comments that were unclear/had incorrect math in `directional_information()`. There are also some other changes that were made when I ran the build routine in RStudio.


nehill197 (approved):

nehill197:
merged into PR #133

PR #144, opened 2026-07-17
Fix dev_skewnorm() residual sign to use the mode, not the mean

nehill197:
The deviance magnitude is computed relative to the fitted distribution's mode (mode_sat), since that is provably where the saturated log-likelihood is achieved for a location family. But the residual sign was computed relative to the theoretical mean instead. For skewed shapes, mode != mean, so points between the two got a small deviance (close to the mode) but the wrong sign (below the mean), producing a gap in the residual distribution where no small positive (or negative) residuals could occur. Use mode_sat for the sign too, matching the reference point used for the magnitude.

dev_skewlnorm() inherits the fix since it delegates to dev_skewnorm().

### DIFF
--- a/R/dev.R
+++ b/R/dev.R
@@ -389,7 +389,7 @@ dev_skewnorm <- function(x, mean = 0, sd = 1, shape = 0, res = FALSE) {
   if (vld_false(res)) {
     return(dev)
   }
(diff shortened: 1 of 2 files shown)


joethorley (approved):

joethorley:
@nehill197 - do you want to merge this?

PR #145, opened 2026-08-10
Add multinomial distribution functions

nehill197:
## Summary
- Add `dev_multinom()`, `log_lik_multinom()`, `ran_multinom()`, and `res_multinom()` for long-format multinomial data (one row per category per trial), using a `group` argument to identify trial membership.
- Validate `group` consistently across all four functions: no `NA`, at least 2 rows per trial, matching `size` and `prob` summing to 1 within a trial, and a consistent row count across groups (catches the common mistake of a `new_expr` being evaluated row-by-row instead of vectorized, and rows lost from a trial).

## Test plan
- [x] `devtools::test()` — 3252/3252 passing
- [x] `devtools::check()` — 0 errors, 0 warnings, 0 notes


R/dev.R
```diff
@@ -267,6 +267,41 @@ dev_lnorm <- function(x, meanlog = 0, sdlog = 1, res = FALSE) {
   dev_norm(log(x), mean = meanlog, sd = sdlog, res = res)
 }
 
+#' Multinomial Deviances
+#'
+#' Models the counts across two or more mutually exclusive categories from a
+#' fixed number of trials, in \emph{long} format: one row per category per
```
joethorley, line 273: In the binomial the number of trials is the number of coin tosses by definition its fixed for a data set. I think the word "fixed" is confusing in this context and should be dropped?

joethorley (approved):
This seems good to me - I've update the documentation and added some more tests in a couple of additional stacked PRs

R/dev.R
nehill197: I think it's the same case for the multinomial! The size - number of trials - is fixed, there are just more possibilities for the number of outcomes than the binomial.

Issue #159, opened 2026-09-09
pskewnorm()/pskewlnorm() use an unnecessary mapply over the already-vectorised sn::psn


StefanoMezzini:
## Problem

`pskewnorm` (`R/skewnorm.R:63`) and `pskewlnorm` (`R/skewlnorm.R:74`) call `sn::psn` through `mapply`:

```r
p <- mapply(sn::psn, x = q, xi = mean, omega = sd, alpha = shape)
```

`sn::psn` is already vectorised over its parameters, so the `mapply` is unnecessary and roughly doubles the cost:

```r
sn::psn(c(0, 0, 0), 0, 1, c(-2, 0, 2))
#> 0.8524164 0.5000000 0.1475836     # correct, no mapply needed
```

## Benchmark

```r
z <- rnorm(2e4)
system.time(pskewnorm(z, 0, 1, 2))   #> 0.661 s
system.time(sn::psn(z, 0, 1, 2))     #> 0.329 s
system.time(pnorm(z))                #> 0.001 s
```

## Note

`qskewnorm`/`qskewlnorm` genuinely do need the `mapply` — `sn::qsn` is not vectorised over `alpha`:

```r
sn::qsn(c(0.5, 0.5), 0, 1, c(-2, 2))
#> Error: the condition has length > 1
```

`sn::dsn` is vectorised, and `dskewnorm` already calls it directly.

So this is only about the two `p*` functions. Switching them to a direct `sn::psn` call would halve their runtime and, as a side benefit, remove the length-recycling surface described in the NA/length-validation issue.

`sn::psn` is still ~300x slower than `pnorm` — that's inherent to the numerical integration in `sn` and not something this package can fix — so this is a modest win, not a transformative one.

StefanoMezzini:
The issue above (which I created as part of a batch using @claude), has some errors. The functions are inconsistent in their use of parallelization:

```r
# psn() is not vectorized over alpha
sn::psn(x = 1:2, xi = 0, omega = 1, alpha = 1) # works
sn::psn(x = 1, xi = 0:1, omega = 1, alpha = 1) # works
sn::psn(x = 1, xi = 0, omega = 1:2, alpha = 1) # works
sn::psn(x = 1, xi = 0, omega = 1, alpha = 1:2) # only one value returned
mapply(sn::psn, x = 1, xi = 0, omega = 1, alpha = 1:2) # returns two values

# qsn() is not vectorized over xi, omega, or alpha
[...]

Issue #158, opened 2026-09-09
sens_skewnorm()/sens_skewlnorm() docs describe 'shape' as non-negative, but it is unbounded


StefanoMezzini:
## Problem

`R/params.R:46` documents `shape` as non-negative:

```
#' @param shape A non-negative numeric vector of shape.
```

For the skew-normal, `shape` (`alpha` in `sn`) is unbounded — negative values give left skew and are fully supported.

`R/skewnorm.R`, `R/skewlnorm.R`, `R/dev.R` and `R/res.R` all override this with a local `@param shape` so their help pages are fine. `R/sens.R` uses `@inheritParams params` only, so the incorrect text surfaces in the rendered docs:

- `man/sens_skewnorm.Rd:14` — `\item{shape}{A non-negative numeric vector of shape.}`
- `man/sens_skewlnorm.Rd:15` — same

Both pages then use negative values in their own examples:

```r
sens_skewnorm(10, 3, -1, 2)     # man/sens_skewnorm.Rd
```

and `man/quant_skewnorm.Rd` / `man/prob_skewnorm.Rd` likewise demonstrate `shape = -2`.

## Suggestion

Add a local `@param shape A numeric vector of shape.` to `sens_skewnorm()` and `sens_skewlnorm()`, matching what `dev_skewnorm()` and `res_skewnorm()` already do.

Worth also checking whether the `params.R` wording is correct for whichever distribution originally motivated it (the beta and gamma families use `alpha`/`beta` and `shape` differently), since if nothing else needs "non-negative" the shared entry could just be relaxed.

Issue #157, opened 2026-09-09
Document that skewnorm 'mean'/'sd' are the location and scale, not the mean and sd


StefanoMezzini:
## Problem

Throughout the skew-normal and skew-lognormal functions, `mean`/`sd` (and `meanlog`/`sdlog`) are passed to `sn` as `xi` and `omega` — the **direct parameters**, i.e. location and scale. For the skew-normal these are not the mean and standard deviation unless `shape = 0`:

- `E[X] = xi + omega * delta * sqrt(2 / pi)`, where `delta = shape / sqrt(1 + shape^2)`
- `SD[X] = omega * sqrt(1 - 2 * delta^2 / pi)`

But `R/params.R` documents them as the moments:

```
#' @param mean A numeric vector of the means.
#' @param sd A non-negative numeric vector of the standard deviations.
```

## Reproduction

```r
set.seed(1)
x <- rskewnorm(2e6, mean = 0, sd = 1, shape = 5)
mean(x)   #> 0.7826015     (documented as 0)
sd(x)     #> 0.6236343     (documented as 1)
```

## Scope

This is a **documentation** problem, not a computational one. The internals convert correctly everywhere I checked — that's what the `sqrt(2 / pi)` terms in `dev_skewnorm`, `res_skewnorm`, `res_skewlnorm`, `sens_skewnorm` and `sens_skewlnorm` are doing, and I verified `sens_skewnorm`/`sens_skewlnorm` preserve the true mean and scale the true sd by `sd_mult` to 4 dp.

The risk is entirely user-facing: anyone simulating data with `ran_skewnorm()`, setting priors, or interpreting fitted `mean`/`sd` parameters will be off by `sd * delta * sqrt(2 / pi)` in location and a factor of `sqrt(1 - 2 * delta^2 / pi)` in spread.

## Suggestion

Keeping the argument names for consistency with the other distribution families in the package seems reasonable. But `?skewnorm` and `?skewlnorm` should state plainly that these are the location and scale (`xi`/`omega` in `sn` terms), give the two moment formulas above, and note that they coincide with the mean and sd only at `shape = 0`. The `@param` entries inherited from `params.R` should be overridden for these functions rather than describing them as moments.

Issue #156, opened 2026-09-09
dskewlnorm() silently ignores non-logical 'log' values; dskewnorm() does not


StefanoMezzini:
## Problem

`dskewnorm` passes `log` straight through to `sn::dsn`, but `dskewlnorm` tests it with `isTRUE()` (`R/skewlnorm.R:48`):

```r
lik <- if (isTRUE(log)) log_lik else exp(log_lik)
```

`isTRUE()` is `FALSE` for anything that isn't a length-1 logical `TRUE`, so numeric and character values silently fall through to the non-log branch instead of being coerced or rejected.

## Reproduction

```r
dskewnorm(1, 0, 1, 2, log = 1)     #> -0.7488043   (log density)
dskewlnorm(1, 0, 1, 2, log = 1)    #>  0.3989423   (NOT the log density)

dskewnorm(1, 0, 1, 2, log = "a")   #> Error: argument is not interpretable as logical
dskewlnorm(1, 0, 1, 2, log = "a")  #>  0.3989423   (silently ignored)

dnorm(1, log = "a")                #> -1.418939 + coercion warning  (base R reference)
```

The silent `log = "a"` case is the worst of these — a typo'd argument produces a plausible-looking number with no signal.

## Suggestion

Validate `log` explicitly in both functions (e.g. `chk_flag(log)`) so the two agree with each other and fail loudly on garbage input, then use the validated flag. `dskewlnorm` needs to keep computing on the log scale internally for the Jacobian, so the `if`/`exp()` structure can stay — it's just the test that should change.

StefanoMezzini:
leaving this for @nehill197 because I'm not sure what the preference is. I'm happy to tackle it if necessary

PR #161, opened 2026-09-09
Fix raw and standardized residuals in res_skewnorm() (#152, #153)

### LINKED ISSUE #152: res_skewnorm(type = "raw") adds the skew correction instead of subtracting it
## Problem

`R/res.R:323`:

```r
raw = x - mean + sd * (shape / sqrt(1 + shape^2)) * sqrt(2 / pi),
```

Operator precedence makes this `(x - mean) + correction`, but the raw residual is `x - E[X]` where `E[X] = mean + sd * delta * sqrt(2 / pi)`. So the skew correction is **added** instead of subtracted, and the residual is off by `2 * sd * delta * sqrt(2 / pi)`.

Note that the `standardized` branch on the very next line parenthesises correctly, so the two branches currently disagree with each other.

## Reproduction

```r
m <- 0; s <- 1; a <- 5
d <- a / sqrt(1 + a^2)
mu <- m + s * d * sqrt(2 / pi)   # 0.7823902
xx <- c(-1, 0, 1, 2)

res_skewnorm(xx, m, s, a, type = "raw")
#> -0.2176098  0.7823902  1.7823900  2.7823900
xx - mu                          # correct
#> -1.7823902 -0.7823902  0.2176098  1.2176098
```

Simulated raw residuals should be centred on zero but are not:

```r
set.seed(1)
xs <- rskewnorm(1e5, 0, 1, 5)
mean(res_skewnorm(xs, 0, 1, 5, type = "raw"))
#> 1.567335      # expected ~0
```

## Fix

```r
raw = x - (mean + sd * (shape / sqrt(1 + shape^2)) * sqrt(2 / pi)),
```

## Tests

The current behaviour is locked in by regression tests that will need updating, e.g. `tests/testthat/test-res.R:1052`:

```r
expect_equal(res_skewnorm(0, 2, 0.5, 10, type = "raw"), -1.60303759425339)
```

The correct value is `-2.39696240574661`.

Worth adding a `shape = 0` equivalence test against `res_norm(..., type = "raw")`, and a check that `mean(res_skewnorm(x, ..., type = "raw"))` is approximately zero for simulated data.

### LINKED ISSUE #153: res_skewnorm(type = "standardized") divides by the variance instead of the standard deviation
## Problem

`R/res.R:324-325`:

```r
standardized = (x - (mean + sd * (shape / sqrt(1 + shape^2)) * sqrt(2 / pi))) /
  (sd^2 * (1 - ((2 * (shape / sqrt(1 + shape^2))^2) / pi))),
```

The denominator is `Var(X)`, not `SD(X)`. It needs a `sqrt()`.

The numerator is correct — this is only about the divisor.

## Reproduction

Two independent confirmations.

**It does not reduce to the normal case.** At `shape = 0` the skew-normal is exactly `N(mean, sd)`, so this must agree with `res_norm`:

```r
res_skewnorm(c(-1, 0, 1, 2), 1, 2, 0, type = "standardized")
#> -0.50 -0.25  0.00  0.25
res_norm(c(-1, 0, 1, 2), 1, 2, type = "standardized")
#> -1.00 -0.50  0.00  0.50
```

Off by a factor of `sd = 2`, exactly as expected from dividing by `sd^2` instead of `sd`.

**Standardized residuals do not have unit variance:**

```r
set.seed(1)
xs <- rskewnorm(1e5, 0, 1, 5)
sd(res_skewnorm(xs, 0, 1, 5, type = "standardized"))
#> 1.612745      # expected ~1
```

For reference, `res_skewlnorm` gets this right (it computes `sd_sln <- sqrt(m2_sln - mean_sln^2)`), and its standardized residuals do have sd ~1.01.

## Fix

```r
standardized = (x - (mean + sd * (shape / sqrt(1 + shape^2)) * sqrt(2 / pi))) /
  sqrt(sd^2 * (1 - ((2 * (shape / sqrt(1 + shape^2))^2) / pi))),
```

## Tests

`res_skewnorm(..., type = "standardized")` currently has **no numeric coverage** — the only tests are NA-propagation checks at `tests/testthat/test-res.R:1109-1125`, which is why this went unnoticed. Suggested additions:

- equivalence with `res_norm(..., type = "standardized")` at `shape = 0`
- `sd()` of standardized residuals ~1 on simulated data

StefanoMezzini:
Fixes two independent bugs in `res_skewnorm()` that were found while auditing the skew normal family. Both are one-line changes to `R/res.R`; the bulk of the diff is test updates.

**This PR was written by Claude Code**, working from an audit of the skew normal implementation. @StefanoMezzini is a co-author on both fix commits and reviewed the work.

## Closes #152 — `type = "raw"` added the skew correction instead of subtracting it

Missing parentheses meant operator precedence parsed the expression as `(x - mean) + correction` rather than `x - (mean + correction)`, so raw residuals were biased by `2 * sd * delta * sqrt(2 / pi)`:

```r
set.seed(1)
xs <- rskewnorm(1e5, 0, 1, 5)
mean(res_skewnorm(xs, 0, 1, 5, type = "raw"))
#> before: 1.567335
#> after:  0.00119711
```

## Closes #153 — `type = "standardized"` divided by the variance instead of the sd

The denominator was `sd^2 * (1 - 2 * delta^2 / pi)`, which is `Var(X)`, not `SD(X)`. Standardized residuals therefore did not have unit variance, and the function did not reduce to `res_norm()` at `shape = 0`:

```r
res_skewnorm(c(-1, 0, 1, 2), 1, 2, 0, type = "standardized")
#> before: -0.50 -0.25 0.00 0.25
#> after:  -1.00 -0.50 0.00 0.50
res_norm(c(-1, 0, 1, 2), 1, 2, type = "standardized")
#>         -1.00 -0.50 0.00 0.50
```

Over 10,000 simulated draws, `sd()` of the standardized residuals goes from 3.34 to 0.99.

## On the test changes

Seven existing expectations encoded the old behaviour and are updated here. That is most of the diff, so it is worth being explicit about why they were wrong rather than the code:

- The `raw` tests that exercised `shape = 0` still pass unchanged, because the skew correction vanishes there. Only the non-zero-`shape` cases moved.
- The `standardized` bug was invisible at `sd = 1`, where the variance and standard deviation coincide, and most surrounding tests use `sd = 1`. The one test that did use `sd = 0.5` asserted exactly double the correct value.

Each fix adds a `test_that()` block that would have caught its bug: equivalence with the corresponding `res_norm()` type at `shape = 0`, an explicit check of the centring and scaling terms, and a distributional check (mean ≈ 0 for `raw`, sd ≈ 1 for `standardized`) on simulated data.

`res_skewlnorm()` was already correct on both counts and is untouched.

## Also included

`5e373b02` (`update {roxygen2} to 8.1.0.9000`) is an unrelated `DESCRIPTION` version bump that was already on this branch, authored by @StefanoMezzini.

## Verification

Full test suite: 3230 passing, 0 failures, 19 pre-existing skips (`extraDistr` not installed locally, plus `skip_on_cran` in `test-sens.R`).

## Not addressed here

Six further issues from the same audit remain open and are out of scope for this PR: #154, #155, #156, #157, #158, #159.

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -53,4 +53,4 @@ Config/testthat/edition: 3
 Encoding: UTF-8
 Language: en-US
 Roxygen: list(markdown = TRUE)
-Config/roxygen2/version: 8.0.0.9000
+Config/roxygen2/version: 8.1.0.9000
--- a/R/res.R
+++ b/R/res.R
@@ -467,7 +467,7 @@ res_skewnorm <- function(
   switch(
     type,
     data = x,
-    raw = x - mean + sd * (shape / sqrt(1 + shape^2)) * sqrt(2 / pi),
+    raw = x - (mean + sd * (shape / sqrt(1 + shape^2)) * sqrt(2 / pi)),
     standardized = (x -
       (mean + sd * (shape / sqrt(1 + shape^2)) * sqrt(2 / pi))) /
       (sd^2 * (1 - ((2 * (shape / sqrt(1 + shape^2))^2) / pi))),
--- a/R/res.R
+++ b/R/res.R
@@ -470,7 +470,7 @@ res_skewnorm <- function(
     raw = x - (mean + sd * (shape / sqrt(1 + shape^2)) * sqrt(2 / pi)),
     standardized = (x -
       (mean + sd * (shape / sqrt(1 + shape^2)) * sqrt(2 / pi))) /
-      (sd^2 * (1 - ((2 * (shape / sqrt(1 + shape^2))^2) / pi))),
+      sqrt(sd^2 * (1 - ((2 * (shape / sqrt(1 + shape^2))^2) / pi))),
     dev = dev_skewnorm(x, mean = mean, sd = sd, shape = shape, res = TRUE),
     chk_subset(x, c("data", "raw", "dev", "standardized"))
   )
--- a/tests/testthat/test-res.R
+++ b/tests/testthat/test-res.R
@@ -1380,7 +1380,7 @@ test_that("res_skewnorm", {
         simulate = TRUE,
         type = "standardized"
       ),
-      c(-0.713119218167245, -0.0574564315318186)
+      c(-0.249771006896373, -0.020124195773685)
     )
   })
   expect_equal(
@@ -1423,7 +1423,7 @@ test_that("res_skewnorm", {
         simulate = TRUE,
         type = "standardized"
       ),
-      c(1.10492371083828, 0.428718918086851)
+      c(0.552461855419138, 0.214359459043425)
     )
   })
   withr::with_seed(101, {
@@ -1435,8 +1435,8 @@ test_that("res_skewnorm", {
       simulate = TRUE,
       type = "standardized"
     )
-    expect_equal(mean(res), 0.0236707225149864)
-    expect_equal(sd(res), 3.34458775450197)
+    expect_equal(mean(res), 0.0070137399798361)
+    expect_equal(sd(res), 0.991016173459387)
   })
   expect_error(res_skewnorm(10, type = "unknown"))
 })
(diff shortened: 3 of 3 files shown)


nehill197 (approved):
Sweet, thanks for fixing this!

